PR #14, opened 2017-03-27
Changed load step to return the result of run()

vitalii-vitrenko:
For example I have the following files:
properties.groovy:
```
map = [
    prop1: 1,
    prop2: 2
]
return map
```

pipeline.Jenkinsfile
```
node ('master') {
    properties= load 'properties.groovy'
}
println properties.prop1
```
The code above works fine in Jenkins pipelines but `loadScript('pipeline.Jenkinsfile')` fails with the error
`groovy.lang.MissingPropertyException: No such property: prop1 for class: properties`
The error occurs because it's assumed that loaded(by `load` step) scripts always `return this`. 
Fixed this issue by returning result of `run()` method from `loadInterceptor`

P. S. Thanks for creating such a great open source testing framework.


ozangunalp:
Thanks for your feedback and this PR. 

You are absolutely right on your diagnosis. I see no problem with returning the result of the script from loadScript. We can go ahead and change the signatures of loadScript and runScript to return object. 

May I ask you several things for this PR :
- Reopen it to the develop branch instead of master. We release the project automatically on push to master. Until we change that we will first merge PRs to develop. 
- Only change the loadScript and runScript methods to return the result of the charged script.
- Add a unit test with your example properties file. 

At least if you can reopen the pull request to develop I can continue on your existing commit. 

Thanks.

ozangunalp:
Great! It seems to do the job. I’m merging the PR. 
I’ll push a commit on top to clean the gimmicks.

Thanks for your contribution.

vitalii-vitrenko:
Thanks for the quick response. 👍

vitalii-vitrenko:
Hello @ozangunalp,
Will these changes be added to the release in the near future?

ozangunalp:
Hi @vitalii-vitrenko, 

Yes I am planning to make a release in this week.

Note that the snapshot artifact is built regularly and available in the OSS Sonatype snapshots repository.

Here is the link
https://oss.sonatype.org/content/repositories/snapshots/com/lesfurets/jenkins-pipeline-unit/1.0-SNAPSHOT/

Issue #15, opened 2017-03-28
Unable to test pipelines using Global Libraries 


joecarlyon:
I have many pipelines that include a global library `@Library('pipelineUtilities@0.5.0')`
We wrap some usage, such as Docker support. The pipelines use that in their code. 
In the unit test, I am unable to overwrite the variables that get used in the library.

Pipeline Code:
```
def unitTestStep(data) {
  node {
    dockerSupport.insideContainer('docker.artifactory.company.com/company/alpine-docker-mvn-ci:0.1.0') {
    echo 'Running with Maven'
    sh "mvn test -P unit-tests -Dparam=${data}"```
  }
}

Unit Test Code:
```void docker_example() {
    binding.setVariable("buildConfiguration", "Sample Build Configuration")
    def script = loadScript("path/to/pipelineCode.groovy")
    script.unitTestStep("This is my data. There are many like it, but this one is mine.")
    printCallStack()
}

Inside of `dockerSupport.insideContainer` there is code being called that uses `buildConfiguration` that is not getting set. The line `binding.setVariable("buildConfiguration", "Sample Build Configuration")` does not override this.

I'm unsure how to mock these things out to move forward.

ozangunalp:
Just to be sure, you did registered your library to the framework as shown in https://github.com/lesfurets/JenkinsPipelineUnit#testing-shared-libraries right?

joecarlyon:
Yeah, I checked that out and it's correct. I may have formed this question wrong, conflating two separate issues I'm facing. This is not my forte.

Global libraries _load_ fine. I'm unable to find the right way to create a typed global variable form an installed plugin https://github.com/jenkinsci/docker-workflow-plugin/blob/master/src/main/resources/org/jenkinsci/plugins/docker/workflow/Docker.groovy (which gets added in https://github.com/jenkinsci/docker-workflow-plugin/blob/master/src/main/java/org/jenkinsci/plugins/docker/workflow/DockerDSL.java )

Also, how do we pass the `org.jenkinsci.plugins.workflow.cps.CpsScript script` (`this` in a `Jenkinsfile`) around?"

ozangunalp:
I recommend you to take a look here : https://jenkins.io/doc/book/pipeline/shared-libraries/#writing-libraries

To resume, 
[...]

PR #16, opened 2017-03-31
TestExampleJob: fix package name and add comment

seanf:
~I also had to change `script.execute()` to `script.run()` for my Java-based test, but for all I know Groovy does something clever with an extension method, so I haven't changed it in the example.~

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,8 @@ return this
 Now using the Jenkins Pipeline Unit you can unit test if it does the job :
 
 ```groovy
-import com.lesfurets.jenkins.helpers.BasePipelineTest
+// Example Groovy test
+import com.lesfurets.jenkins.unit.BasePipelineTest
 
 class TestExampleJob extends BasePipelineTest {
         


seanf:
I worked out why I was having a problem with `script.execute()` vs `script.run()`: I was testing with a `Jenkinsfile`, which seems to be different from other pipeline scripts: it doesn't define an `execute` method (or `return this`), thus there is no `execute` method to call. From Java, if you do want to call an `execute` method, I suppose you have to use reflection, or perhaps `script.evaluate("execute()")` would work.

When unit testing such a `Jenkinsfile`, apparently you shouldn't call `script.run()` either, because you get a `CpsCallableInvocation`. In fact, I see in the code that `loadScript` indirectly calls `script.run()`.

ozangunalp:
Hi @seanf, thanks for your feedback. 
Actually there is already an ongoing PR (#7) which makes the changes you propose for making the package name change in the documentation. 

As for your confusion between `execute` and `run` methods : 
In the current version the `loadScript` method loads, runs and returns the script. 

The `script.execute()` call that you find in the example (and in many tests in the project) is just a call to the method defined inside the loaded script (cf. [exampleJob.jenkins](https://github.com/lesfurets/JenkinsPipelineUnit/blob/master/src/test/jenkins/job/exampleJob.jenkins#L1)). As you noticed the script returns itself with `return this`. If you are unit testing this kind of script, in Groovy you can benefit from dynamic typing and call the method directly, and in Java you need to use reflection (see [TestPipelineJava.java](https://github.com/lesfurets/JenkinsPipelineUnit/blob/master/src/test/java/com/lesfurets/jenkins/TestPipelineJava.java#L35)).

Btw this pattern is pretty useful when you are loading scripts within your jenkins pipelines with `load`. As for the `Jenkinsfile`, it is always called by Jenkins itself, that is why you (usually) don't do any `return this` etc.

In the next release, as discussed in #14,  the `loadScript` method will load, run and return the returning value of the script. So this will be the script itself if loaded script ends with `return this`, or else whatever object that the script returns. But after the discussions in #15, it may be a better idea to separate `loadScript` and `runScript`, I will try to do that without breaking existing `loadScript` calls. Providing a better wrapped `runScript` method should avoid the `CpsCallableInvocation` exception you are receiving.

ozangunalp:
I hope that helps you. I am closing this PR because it is the duplicate of #7.

seanf:
Thanks for the explanation @ozangunalp .

PR #18, opened 2017-03-31
Bring back failing test

seanf:


ozangunalp:
So in that test class we try to see if the testing framework is able to simulate and thus detect common script errors in Jenkins. 

The test in question is linked to the [JENKINS-26481](https://issues.jenkins-ci.org/browse/JENKINS-26481) issue which causes an `UnsupportedOperationException` when using DefaultGroovyMethods. This test expects that exception (not the ComparisonFailure) which is not yet properly simulated by the framework. This is why it is ignored for now.

seanf:
Oh, I see now that the behaviour in [JENKINS-26481](https://issues.jenkins-ci.org/browse/JENKINS-26481) has changed since it was first reported (when it just executed the closure *once*, like my test). The `UnsupportedOperationException` must have been introduced afterwards, since its message actually includes a reference to JENKINS-26481.

Even if pipeline-unit can't simulate the `UnsupportedOperationException` yet, at least CPS mode doesn't execute the closure multiple times, so it can still provide a little help in catching these script problems.

Issue #57, opened 2017-08-30
Utils.groovy has @Grab depedency which breaks build inside firewall


sic-bre:
The Utils class (src\test\resources\libs\commons@master\src\net\courtanet\jenkins\Utils.groovy) that is used in unit tests for the Shared Libraries uses a @Grab dependency that makes it unnecessarily difficult to reproduce the build behind a firewall without changing the checked in code or adding cumbersome configuration options specific to the build.  Since unit tests should be self contained either the script Grab dependency should use a @GrabResolver referencing a jar file that has been downloaded as part of the Gradle build, or the test case example should be rewritten to not need an external dependency (which in this case is not significant to the JenkinsPipelineUnit functionality under test.)

ozangunalp:
I agree that this test is far from ideal. It should not download the jar etc. I'll look more into the grab configuration. Feel free to suggest a PR to fix it. 

Thanks

sic-bre:
Actually there is a much simpler solution, just add it to the testCompile configuration and it passes the unit tests:

Remove from Utils.groovy:
        @Grab('org.apache.commons:commons-math3:3.4.1')

Add to the top level build.gradle in dependencies {} block:
        testCompile group: 'org.apache.commons',    name: 'commons-math3',  version: '3.4.1'

I'll add a PR if you still want it but this is so simple...

ozangunalp:
It is there to verify if the Grab is working as expected. 
The shared library test can be splitted, that's for sure. But I would not remove it.

sic-bre:
Take a look at the PR I added this afternoon I think it does what you want it to do.

ozangunalp:
Thanks for the pr. I saw it and it should work as expected but didn't had time to test it yet. I'd prefer doing it on junit though.

sic-bre:
I don't understand your comment "I'd prefer doing it on junit though."?  My proposed changes setup the local test environment so that the tests which are still in junit are left unchanged.  I don't see how you can do this inside the junit tests themselves as the information about repositories and networking has to be passed to it somehow.  Making the tests work behind a firewall you can't assume anything except that the user has set up for Gradle itself for the local network.  In our case we use a standard init.gradle script to run a plugin in settings.gradle which among other things remaps all the repositories and configures pluginManagment settings.  Other people may choose to set the global java proxy settings but that can't be assumed to work everywhere.

PR #58, opened 2017-08-31
Changes needed to support isolated @Grape() testing

sic-bre:
Adds configuration and copies dependency to local filesystem used by @Grape() to the local filesystem and overrides the defaultGrapeConfig.xml during unit tests to use the local copy.  Currently limited to non-transitive dependencies but enough to make the unit tests run behind a firewall.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,13 @@ group = "com.lesfurets"
 archivesBaseName = "jenkins-pipeline-unit"
 version = "1.2-SNAPSHOT"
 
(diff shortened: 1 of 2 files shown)


sic-bre:
This pull request is for Issue #57

ozangunalp:
@sic-bre
Indeed downloading the grape dependency with gradle seems the way to go.
Your changes are ok for me. I'll move the task you wrote into `buildSrc` directory.

PR #60, opened 2017-09-01
testNonRegression: Generate actual files on error

olivierdagenais:
`testNonRegression` is insanely useful to write characterization tests and then maintain those tests as changes are made subsequently to the `Jenkinsfile`.

The point of this pull request is to add a feature that replaces the need to add this code to my test suites:
```
// uncomment the line below when you want to update the reference callstacks
//System.setProperty(RegressionTestHelper.PIPELINE_STACK_WRITE, "true");
```
...such that if the output of a run changes, I don't have to:
1. Temporarily un-comment the line
2. Re-run the test
3. Re-comment the line

I instead can fire up my diff tool and _copy_ the changes from the newly-generated `.actual` file into the corresponding suite file (if I want to).

As a bonus, the "expected" file will now be written if one didn't exist before, instead of the corresponding test failing with a `FileNotFoundException`.

Please let me know if there are additional tests you would like me to write/run.

Thanks!

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/RegressionTestHelper.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/RegressionTestHelper.groovy
@@ -32,6 +32,11 @@ class RegressionTestHelper {
         String callStack = helper.callStack.join('\n') + '\n'
         final expected = referenceFile.text.normalize()
         final actual = callStack.normalize()
+        if (!expected.equals(actual)) {
+            final actualFileName = targetFileName + ".actual"
+            final actualFile = new File(actualFileName)
+            writeStackToFile(actualFile, helper)
+        }
         assertThat(actual)
                         .as('If you intended to update the callstack, use JVM parameter -D%s=true', PIPELINE_STACK_WRITE)
                         .isEqualTo(expected)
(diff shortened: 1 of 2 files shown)


ozangunalp:
Thanks for this pr. We realized that issue after the release, but it was too late.  
I’ll take a look as soon as I can.

sic-bre:
Another simpler approach would be to add the following to your build.gradle file:

```
test {
    systemProperty "pipeline.stack.write", System.properties["pipeline.stack.write"]
}
```
Now if you run `gradle -Dpipeline.stack.write=true cleanTest test` it will rewrite the output since the property is getting passed from the gradle java process to the forked test process.  If you only want to refresh a subset of the tests add the `--tests *` flag limiting which tests you are going to run.

EQuincerot:
I would say, `pipeline.stack.write` is supposed to indicate whether the reference files must be written.
When you run the test and the reference file is different, we suggest to load the file.
I would rather suggest to check the file existence just before comparing it:
```
        assertThat(referenceFile.exists())
                .as('To initiate the callstack reference file, user JVM parameter -D%s=true', PIPELINE_STACK_WRITE)
                .isTrue()
```

Issue #73, opened 2017-10-11
Provide a way to cache the PipelineTestHelper and dependent libraries


jacobfederer:
The `PipelineTestHelper ` does a lot of heavy lifting e.g. loading a library from various sources. Since the library content doesn't change during the test execution it would be good to provide a way to cache this helper and it's dependent libraries to speed up the test execution and reducing the I/O effort.

ozangunalp:
You can call the `BasePipelineTest#setUp` method in the per class initializer of your test runner.
For example if you are using JUnit you can call the setUp method in your `@BeforeClass` method as in 

```java
class PipelineJobTest extends BasePipelineTest {
    
    @BeforeClass
    void initTests() throws Exception {
        super.setUp();
    }
```

jacobfederer:
Unfortunately this is not possible since `@BeforeClass` requires the method to be static. This is caused by the fact that every `@Test` method will be executed in a separate instance of the referenced test class.

ssbarnea:
We have our shared library inside a git repository and we are using the `.retriever(GitSource.gitSource(...))` approach to load the shared library... and we have >1300 jobs to tests. If that's true it means that we perform a serious DDOS against git server.

Any workarounds?

stchar:
I assume this one could be closed.

PR #74, opened 2017-10-11
Add HelperSingleton to cache PipelineTestHelper

jacobfederer:
Feature implementation for #73 

Adds a singleton for the `PipelineTestHelper` class. This can especially be used in a `@BeforeClass` static junit method so the singleton instance and dependent libraries will be used during the whole test execution.

### DIFF
--- a/src/test/groovy/com/lesfurets/jenkins/TestHelperSingleton.groovy
+++ b/src/test/groovy/com/lesfurets/jenkins/TestHelperSingleton.groovy
@@ -1,0 +1,59 @@
+package com.lesfurets.jenkins
+
+import com.lesfurets.jenkins.unit.BasePipelineTest
+import com.lesfurets.jenkins.unit.HelperSingleton
+import com.lesfurets.jenkins.unit.PipelineTestHelper
+import org.junit.Before
+import org.junit.BeforeClass
+import org.junit.Test
+
+import static com.lesfurets.jenkins.unit.global.lib.LibraryConfiguration.library
+import static com.lesfurets.jenkins.unit.global.lib.LocalSource.localSource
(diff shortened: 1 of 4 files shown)


fordperfect78:
👍 would be great to get this in

nevskrem:
Could someone please merge?

OliverNocon:
Hi, any update on this? Could this be merged please? 
We are suffering from increasing test times (about 2-3 seconds overhead for each test case due to loading a shared library) and it would be great to see this in the standard.

src/test/groovy/com/lesfurets/jenkins/TestHelperSingleton.groovy
```diff
@@ -0,0 +1,65 @@
...
+
+    @BeforeClass
+    static void beforeClass() {
+        HelperSingleton.singletonInstance = new PipelineTestHelper()
+
+        String sharedLibs = this.class.getResource('/libs').getFile()
+
+        def library = library().name('commons')
+                .defaultVersion("master")
+                .allowOverride(true)
```
stchar: I'm not sure how singleton would support backward compatibility. This may introduce a potential bug if singleton instance altered from previous test class would be still available to next test class. 

It means if we have TestA, TestB chain 
TestA register some method, TestB inherits this method via singleton instance   
If we change an order so TestB goes first it will throw an exception I suppose.

stchar: Moreover this applicable for exactly test case.  
```
// what if I change the order commenting out some tests
// @Test
// void test_first() {}
@Test
void test_second() {}
```

Issue #70, opened 2017-10-05
Question: Testing Shared Library definition using JenkinsPipelineUnit


mkobit:
I (finally) released the first version of https://github.com/mkobit/jenkins-pipeline-shared-libraries-gradle-plugin , which is a Gradle plugin intending to aid in the authoring of shared libraries. I'm sure an additional plugin could be added to support consumers of `JenkinsPipelineUnit` for build pipeline authors that also want to test their pipeline.

I wrote up an example library at https://github.com/mkobit/jenkins-pipeline-shared-library-example, but was struggling in writing a unit test against the local shared library functionality (at https://github.com/mkobit/jenkins-pipeline-shared-library-example/blob/master/test/unit/groovy/com/mkobit/libraryexample/JPUExampleSpec.groovy right now).

Any suggestions or example  to look at for using this as a test harness for Shared Library definition?

jimcroft:
Hi @mkobit 

I've put a basic skeleton example of what I'm doing here: https://github.com/jimcroft/jenkinslib-example. Hope it helps. If that example does help you load your Jenkins library from a local location (rather than an scm source) could you add a comment to https://github.com/jenkinsci/JenkinsPipelineUnit/pull/75?

Thanks

Jim

balux:
Waiting for "local-location" solution too! Meanwhile you can use a copy task in your `build.gradle`
```groovy
task initLibraries(type: Copy){
    from('./') {
        include 'src/**/*'
        include 'vars/**/*'
    }
    into 'build/classes/groovy/test/nextlayerci-example@master'
}

test.dependsOn initLibraries
```
https://github.com/nextlayer-ops/jenkins-pipeline-shared-library-example/blob/master/build.gradle#L12-L20
and 
https://github.com/nextlayer-ops/jenkins-pipeline-shared-library-example/blob/master/test/groovy/SayHelloTest.groovy#L18-L30

Btw. If you want to use maven you can use the same trick:
https://github.com/nextlayer-ops/jenkins-pipeline-shared-library-example-maven/blob/master/pom.xml#L54-L80

mkobit:
https://github.com/jenkinsci/JenkinsPipelineUnit/pull/75 and https://github.com/jenkinsci/JenkinsPipelineUnit/pull/64 are both related to this, would be great to see one of those move forward

balux:
Meanwhile I've found a better solution. Thanks to the SAP guys ;)

[...]

PR #75, opened 2017-10-13
Update library loading so branch not required for localSource locations.

jimcroft:
I was trying to use JenkinsPipelineUnit to test a library I'm writing. I want to use this from within local 
copy (clone) of the repo using the localSource retriever. This was problematic as the code required
setting defaultVersion (branch) which it then appended to the path to the library root.

This PR allows the following usage:

```
        def library = library()
                .name('jenkins-library')
                .allowOverride(true)
                .implicit(false)
                .targetPath(dirPath)
                .retriever(localSource(dirPath))
                .build()
```
where `dirPath` is the project root, without `@<branch>` or other version identifier required in the filesystem path.

I'm not a Java or Groovy coder so this PR probably has issues so any feedback welcome (even if it's just pointing out that I'm not using it properly! :) )

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/global/lib/LibraryConfiguration.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/global/lib/LibraryConfiguration.groovy
@@ -18,7 +18,7 @@ class LibraryConfiguration {
     String targetPath
 
     LibraryConfiguration validate() {
-        if (name && defaultVersion && retriever && targetPath)
+        if (name && retriever && targetPath && ((retriever instanceof LocalSource) || defaultVersion))
             return this
         throw new IllegalStateException("LibraryConfiguration is not properly initialized ${this.toString()}")
     }
--- a/src/main/groovy/com/lesfurets/jenkins/unit/global/lib/LocalSource.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/global/lib/LocalSource.groovy
@@ -11,7 +11,14 @@ class LocalSource implements SourceRetriever {
 
     @Override
     List<URL> retrieve(String repository, String branch, String targetPath) {
-        def sourceDir = new File(sourceURL).toPath().resolve("$repository@$branch").toFile()
+        def sourceURLPath = new File(sourceURL).toPath()
+        def sourceDir
+        if (branch) {
+            sourceDir = sourceURLPath.resolve("$repository@$branch").toFile()
+        } else {
+            sourceDir = sourceURLPath.resolve(repository).toFile()
+        }
+
         if (sourceDir.exists()) {
             return [sourceDir.toURI().toURL()]
         }


mkobit:
For clarification, you want to use JenkinsPipelineUnit to test a shared library? If so, this is also related to my question in https://github.com/jenkinsci/JenkinsPipelineUnit/issues/70

marcusholl:
Sounds to be related to #64

oliverlockwood:
FWIW, we (before finding this PR) have locally implemented a similar solution to the same problem.

Is there any chance of the project maintainers taking this on board, or providing feedback so it could be moved forward?

brianeray:
This would be great. I hacked up a solution involving a temporary symbolic link to the project in the form `libName@master`  as described in this [#64 comment](https://github.com/jenkinsci/JenkinsPipelineUnit/pull/64#issuecomment-364197979). But it would be better to have an official upstream solution.

giorgiosironi:
:+1: for this, I have been struggling for months intermittently to add tests to my organization-wide shared library. If a commit breaks a step, it can impact dozens of builds.

src/main/groovy/com/lesfurets/jenkins/unit/global/lib/LibraryConfiguration.groovy
```diff
@@ -18,7 +18,7 @@ class LibraryConfiguration {
     String targetPath
 
     LibraryConfiguration validate() {
-        if (name && defaultVersion && retriever && targetPath)
+        if (name && retriever && targetPath && ((retriever instanceof LocalSource) || defaultVersion))
```
robertoaloi: I believe this class should be agnostic of the retriever's "type".

src/main/groovy/com/lesfurets/jenkins/unit/global/lib/LocalSource.groovy
```diff
@@ -11,7 +11,14 @@ class LocalSource implements SourceRetriever {
 
     @Override
     List<URL> retrieve(String repository, String branch, String targetPath) {
```
robertoaloi: The real problem here is that the (abstract) concept of _SourceRetriever_ is too tightly coupled with the one of a (git) repository and that we are bending the implementation accordingly. Rather than always requiring a branch and ignoring it when it does not make sense (i.e. in line 18), it would be much better if the `retrieve` API had a more generic signature. For example, it could take take a single map instead of the `repository/branch` tuple. This would allow us to specialize the map according to the needs of the specific retriever.

msmagoo87:
Hoping this can be merged soon. I'm using the jenkins-pipeline-unit to test a shared library via maven instead of gradle and am running into this issue with no real way to work around it.

mkobit:
Any thoughts around merging something like this or https://github.com/jenkinsci/JenkinsPipelineUnit/pull/64 to support the "local source code" functionality?

stchar:
@mkobit,
@ozangunalp  has already merged #64 instead, 
I verified it locally with @jimcroft test repo

I suppose this one could be canceled

PR #86, opened 2018-01-16
use Java 8.  Tests won't run from IntelliJ without using Java 8

tweissin:


smola:
@tweissin There's probably no problem with this change, but tests run correctly in IntelliJ 2017.3.3. At least if imported properly as a Gradle project.

PR #89, opened 2018-02-07
Zap file descriptor leak in doLoadLibrary() stemming from the Directo…

brianeray:
…ryStream returned by `Files.list(Path)`. Had to resort to an explicit `close()` because the following didn't work:

- The try-with-resources pattern. This results in valid syntax for Java but invalid syntax for Groovy.
- Groovy's `Closeable.withCloseable(Closure)`. This *should* have worked, in fact it is the recommended equivalent to the try-with-resources pattern previous. For some reason I couldn't get it to compile statically. With `@CompileStatic(TypeCheckingMode.SKIP)` it compiled but then blew up at runtime with some weird exception.

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/global/lib/LibraryLoader.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/global/lib/LibraryLoader.groovy
@@ -87,17 +87,19 @@ class LibraryLoader {
                 groovyClassLoader.addURL(resourcesPath.toUri().toURL())
 
                 if (varsPath.toFile().exists()) {
-                    Files.list(varsPath)
-                                    .map { it.toFile() }
-                                    .filter ({File it -> it.name.endsWith('.groovy') } as Predicate<File>)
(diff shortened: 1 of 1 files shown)


brianeray:
I should add a test for this ... I tested it manually, locally in one of my Pipeline projects that kills the temporary directory holding the lib dependendencies resolved from a Git repo. This happens in my project's Spock `cleanupSpec()` phase.

The fd leak prevented the deletion of the lib dependency directories (but not the files therein) and the temp workspace directory above it. It was not even possible to hack around it in my Spock spec via `File.deleteOnExit()`.

FWIW I hit this on Windows 8.1 using Gradle 4.4.1 running on a Zulu OpenJDK 1.8.0_152 with `groovy-all:2.4.13:indy`, `spock-core:1.1-groovy-2.4`, `jenkins-pipeline-unit:1.1`, and a custom implementation of `SourceRetriever` using `jgit:4.9.2.201712150930-r` instead of locally installed Git.

PR #109, opened 2018-07-04
Add partial support of loading a shared library dynamically

stchar:


svargasConnexta:
This PR would be EXTREMELY beneficial for those of us who need to use private libraries (@Library annotation in Jenkins only works with public libraries)

Issue #51, opened 2017-07-26
Testing Shared Libraries that use Jenkins commands


vle-tom:
I have been trying to test a shared library and went the route of wrapping it in a test pipeline and using this framework to exercise it. The library has a method that looks like 

`def getBuildProperties(){ return readFile('file.properties')}` where readFile is a Jenkins command. 

My test case is set up to register the method like so, 
`	helper.registerAllowedMethod(method("readFile", String.class), { file ->
	    return FileUtils.readFileToString(new File(file), Charset.forName("UTF-8"))
	})
`

The pipeline looks like
`
def execute() {
	def util = new Util()

	node() {
	
       stage('test') {
         readFile('file.properties') // works
         util.getBuildProperties() // does not work
      }
  }
}
`

The error I get is, 
`groovy.lang.MissingMethodException: No signature of method: com.company.Util.readFile() is applicable for argument types: (java.lang.String) values: [file.properties]
...
`

Is there something I'm missing here or is this not supported?
Is there a mechanism for testing a shared library in a standard unit test with the ability to have the Jenkins commands available, mocked or otherwise?

headcrabmeat:
Hey, you forgot some lines in the pipeline
```
@Library('le_library')
import package.path.Util
```
Take a look on this one https://github.com/headcrabmeat/pipeline-sharedlib-testharness

headcrabmeat:
Hi, I read it again =). 
I think that you missed a reference to pipeline object in your `getBuildProperties()` 
That it is why the mock works with the pipeline but not with the library
see https://jenkins.io/doc/book/pipeline/shared-libraries/#accessing-steps

so it should look like 
```
// library class
class Util ... {
  Util(script) {
    this.script =script
  }

  def getBuildProperties(){ 
   return script.readFile('file.properties')
  }

```

```
// pipeline
def execute() {
	def util = new Util(this)

	node() {
	
        stage('test') {
          readFile('file.properties') // works
          util.getBuildProperties() // should work now )))
        }
    }
}
```

attiand:
I think @vle-tom was trying to access steps outside of the scope of an enclosing class, i.e. the first example (`src/org/foo/Zot.groovy`) in https://jenkins.io/doc/book/pipeline/shared-libraries/#accessing-steps 

i.e. 
```groovy
// src/com/company/Util.groovy
package com/company;

def getBuildProperties(){ 
   return script.readFile('file.properties')
}
```
[...]

PR #111, opened 2018-07-12
Fixed intercepting GCL for Global Shared Libraries

radekl:
There is an issue with Intercepting GCL. The parseClass method is never executed thus making libraries attached by e.g.: `helper.registerSharedLibrary(library)` not able to execute Jenkins native methods (ones registered by `registerAllowedMethod` etc). 

This PR fixes the issue and allows to test the global shared libraries locally.

I guess it also fixes issue #51.

P.S. I haven't removed the old parseClass as I'm not sure if it is not executing all the time or only for this particular case.

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
@@ -16,7 +16,7 @@ class InterceptingGCL extends GroovyClassLoader {
 
     @Override
     Class parseClass(GroovyCodeSource codeSource, boolean shouldCacheSource)
-                    throws CompilationFailedException {
+            throws CompilationFailedException {
         Class clazz = super.parseClass(codeSource, shouldCacheSource)
         clazz.metaClass.invokeMethod = helper.getMethodInterceptor()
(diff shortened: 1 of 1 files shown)


radekl:
Although it works with Allowed Methods it doesn't if the shared library inter-uses methods. Let say that we have following library:
```
vars/
- utils.groovy
- commands.groovy
```
both `utils` and `commands` can use methods registered as allowed methods, but they cannot call methods from themselves. E.g.: `commands.groovy` cannot use `utils.someUtilMethod()`

radekl:
Fixed above with some kind of a dirty hack in a9680fc. I have overwritten loadClass method to forcibly execute `recompile` which executes indirectly the `parseClass` method. Works perfectly for my case where I need to test Shared Libraries.

radekl:
Fixed tests

stchar:
@erikzaadi, sorry for a long response, may I ask you to write a test case.

s1apped:
Hi. any news about it?

stchar:
@s1apped, the pull request needs rebasing and some tests, any help is welcome.

radekl:
Unfortunately I won't be able to participate with this PR anymore. Please feel free to add tests if you can.

stchar:
work is continued in #193

PR #117, opened 2018-08-31
registerMethodCall in PipelineTestHelper should clone the method args

macg33zr:
registerMethodCall in PipelineTestHelper should clone the method arg as they may change later in the pipeline execution. This fixes an issue with incorrect call stack recordings that show the variables passed to a step in their state at the end of the pipeline execution (e.g. global vars to the pipeline). Also added a test example to exercise this.

I also have an example pipeline here that shows the issue:
https://github.com/macg33zr/pipelineUnit/blob/master/exampleJobs/globalVariable/Jenkinsfile

### DIFF
--- a/src/test/resources/callstacks/TestRegressionGlobalVar_globalVar.txt
+++ b/src/test/resources/callstacks/TestRegressionGlobalVar_globalVar.txt
@@ -1,0 +1,8 @@
+   globalVar.run()
+      globalVar.node(groovy.lang.Closure)
+         globalVar.stage(One, groovy.lang.Closure)
+            globalVar.echo(Stage One)
+            globalVar.doWithProperties({PROP_1=VAL_1})
+         globalVar.stage(Two, groovy.lang.Closure)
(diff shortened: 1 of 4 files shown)


src/test/resources/callstacks/TestRegressionGlobalVar_globalVar.txt
```diff
@@ -0,0 +1,8 @@
+   globalVar.run()
+      globalVar.node(groovy.lang.Closure)
+         globalVar.stage(One, groovy.lang.Closure)
+            globalVar.echo(Stage One)
+            globalVar.doWithProperties({PROP_1=VAL_1})
```
macg33zr: If you don't apply the code fix, this call stack here in the first stage will show:

`globalVar.doWithProperties({PROP_2=VAL_2, PROP_1=VAL_1})`

At this point in the execution, that would clearly not be correct as PROP_2 is added in the second stage.

stchar:
@macg33zr, merged, thank you

PR #121, opened 2018-10-04
Replace time and object hash before comparing call stack for regression

restfulhead:
Some of our pipelines add the current time to log messages. This causes regression tests to fail, as the time value will always be different. 

We're also sometimes passing objects that don't have a `toString()` method implemented. This also causes regression failures, because the object id changes each time.

This simple change addresses this, by replacing certain patterns with a fixed value, more specifically:
- `(\d{2}:\d{2}:\d+)` -> `%TIME%`
- `([@][0-9a-f]+)` -> `%HASH%`
- `[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}` -> `%UUID%`

Example before:
```
lib-something.echo([MyTest] [08:33:22] [INFO] Hello World)
lib-something.doStuff(mypackage.LogHelper@12f3ab)
```
Example after:
```
lib-something.echo([MyTest] [%TIME%] [INFO] Hello World)
lib-something.doStuff(mypackage.LogHelper%HASH%)
```

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/RegressionTestHelper.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/RegressionTestHelper.groovy
@@ -29,7 +29,7 @@ class RegressionTestHelper {
             writeStackToFile(referenceFile, helper)
         }
 
-        String callStack = prepareCallStack(helper.callStack)
+        String callStack = prepareCallStack(helper.callStack) + System.lineSeparator()
         assertThat(callStack.normalize())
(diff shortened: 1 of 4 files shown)


stchar:
Hi, it looks you're trying to post-edit behavior of java.util.Date.
I believe you can use `groovy.lang.mockFor` instead.
I don't want to criticize your work but replacing values from callStack is not a common approach the library should be.

PR #13, opened 2017-03-22
Support declarative pipeline syntax

ozangunalp:
Support for testing declarative pipeline syntax.

Provides a DeclarativePipelineTest base class that 
- Does syntactic checking of the declarative pipeline model
- Mocks the [ModelInterpreter](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy) provided by the [pipeline-model-definition-plugin](https://github.com/jenkinsci/pipeline-model-definition-plugin) and sets props/params/credentials etc. and calls steps on stages, with when predicates and post actions.

Yet there are several caveats with this approach: 
- It does not verify whether there is any if,else,loop instructions on regular steps blocks, 
- Mock model interpreters should be easily extensible by users, which is not possible for now. For now the only extension point is the `pipeline` keyword, which involves rewriting every model interpreter.
 DeclarativePipelineTest does not use CPS transformation so it does not verify the pipeline against serialization issues.

PR #133, opened 2019-02-10
Add declarative pipeline harness

dgoetsch:


dgoetsch:
example usage:

```
class HarnessTest extends BasePipelineTest {

    String jenkinsfileName = "Jenkinsfile"
    Script jenkinsfile = null

    DeclarativePipelineHarness harness

    def loadJenkinsfile() {
        this.jenkinsfile = super.loadScript(jenkinsfileName)
    }


    @Before
    void setup() {
        if(harness == null || jenkinsfile == null) {
            super.setUp()
            harness = new DeclarativePipelineHarness(this)
            loadJenkinsfile()
        } else {
            harness.clearAllCalls()
        }
    }

    @Test
    void smokeTest() {
        jenkinsfile.run()
        harness.runCall('pipeline')
        harness.runCall('stages')
        harness.runIdentified('stage', 'Build')
        harness.runCall('steps')
        def call = helper.callStack.find { it.methodName == 'sh' }
        assertThat(call.args.toList(), equalTo(['gradle build']))
    }
}
```

Before I formalized this unit test, I would like to get feedback.  Once I am granted tentative approval, I will add a unit test.

dgoetsch:
This approach is very different in `https://github.com/jenkinsci/JenkinsPipelineUnit/pull/13/files`, in that all scopes are not implicitly executed. It forces the user to define which part of the pipeline is being executed, and allows stricter control and verification for complex pipelines.

This has the effect of reducing the coverage of any individual test; which decouples dependencies across disparate blocks (really helps with testing complex pipelines, i'm working with a pipeline with 40+ stages).  This will make tests more maintainable and reduce the number of tests that would need to be changed when the pipeline is changed.

src/main/groovy/com/lesfurets/jenkins/unit/DeclarativePipelineHarness.groovy
```diff
@@ -0,0 +1,164 @@
...
+
+    void runCall(String name, Integer idx = 0, Boolean doClear = true) {
+        CallStack callStack = callStacks[name]
+        if(callStack== null) {
+            throw new RuntimeException("No such call stack ${name}")
+        }
+        callStack.run(idx, doClear)
+    }
+
+    public <T> void runIdentified(String name, T value, Boolean doClear = true) {
```
dgoetsch: This method name sucks; but I couldn't come up with a bette one.

Maybe `runIdentifiedCall` or `runCallById` (which would also play nicely with `runCallByIdx`).

src/main/groovy/com/lesfurets/jenkins/unit/DeclarativePipelineHarness.groovy
```diff
@@ -0,0 +1,164 @@
...
+        String name
+        PipelineTestHelper helper
+
+        List<Closure> calls = []
+
+        CallStack(name, helper) {
+            this.name = name
+            this.helper = helper
+
+            this.helper.registerAllowedMethod(name, [Closure.class], { calls.add(it) })
```
dgoetsch: This is a fundamentally different approach than most of this framework, because the closures in the jenkinsfile are not implicitly executed but must be described by  the test itself. (which imo makes it a more focused test)

dgoetsch:
@stchar of course.  I will update this PR when I have some time

dgoetsch:
@stchar I finally got around to updating this.  I've added a unit test and completed the implementation, as well as added a bit of documentation to the README.

I didn't not add a unit test that tests libraries, could do that if desired.  I dont' know if it would be a valuable test since the implementations are properly isolated.  I have used the same code with libraries and it uses the same mechanism to run as the other tests, just adds some structure around the steps. From an implementation standpoint, the library loading is unaffected by the usage of the BaseDeclarativePipelineTest

dgoetsch:
Ooof. I just realized I did my most recent commit on a different machine and it didn't recognize my git user

src/main/groovy/com/lesfurets/jenkins/unit/DeclarativePipelineHarness.groovy
dgoetsch: renamed to `runClosureStep`

dgoetsch:
@stchar is there anything else that you need from me before this can be merged?

README.md
```diff
@@ -385,24 +385,46 @@ libraryJob.run()
...
+your tests describe which part of the jenkinsfile your test is verifying and don't interract with the rest of the file.  Here is an example that
+verifies that sh was invoked during the "init" stage.
+
+```
+@Test
+void runAStage() {
+    super.jenkinsfile.run()
+    harness
+        .runStage('init')
+        .runClosureStep('steps')
```
AlanFoster: @dgoetsch This looks cool :+1:

Is there a way to just run an entire declarative pipeline in a similar fashion to Jenkins too?

src/main/groovy/com/lesfurets/jenkins/unit/DeclarativePipelineHarness.groovy
```diff
@@ -0,0 +1,157 @@
...
+            if(! (idx < calls.size())) {
+                throw new RuntimeException("not enough calls in stack for ${name}, expected idx: ${idx}, size: ${idx}")
+            }
+            def c = calls[idx]
+            if(doClear) { clearAllCalls() }
+            c()
+        }
+    }
+
+    class ParameterizedCallStack<T> {
```
stchar: `ParameterizedCallStack` and `CallStack` looks almost the same, 
 i'm just wondering if these classes could be coupled by inheritance e.g. 
```
class ParameterizedCallStack extends CallStack
```

stchar: or maybe replace it with just one class, 
As result you can reduce the logic splits you need to handle parameterized or non-parameterized closures.

stchar:
@dgoetsch, I took a closer look on your implementation, it looks cool, it allows to test declarative pipeline running it stage by stage.   
it also has some lack of support of some essential features (imho) of declarative pipeline like:
* `parameters {}` 
* `pre {}` 
* `post {}`
* `environment {}`
There was an attempt to implement so in #13 (which you mentioned)
It has an attempts to implement these features of declarative pipeline. 

Maybe we can try to combine your solution with Ozan's. What do you think?

README.md
dgoetsch: I believe this is relevant to the comment from @stchar asking about built in support for `parameters` / `environment` / `pre` / `post`.

dgoetsch:
I'll wait until we resolve conversations to fix the readme conflicts

stchar:
Sorry to say but this PR has lack of features,  and needs significant rework IMHO. 
I'm canceling it in favour of #13

PR #155, opened 2019-09-19
Add additional allowed methods

pontuslaestadius:
Currently there exists a number of functions and closures which Jenkins Pipeline supports that are not included in BasePipelineTest.
This proposition adds additional allowed methods that allows a wider usage of JenkinsPipelineUnit.


stchar:
thank you

PR #167, opened 2020-01-06
Add missing inputs to the incremental task "test"

theosotr:
This pull request fixes the specification of the gradle task test.
This task depends on some test resources that are located in the `src/test/jenkins` directory.

This commit considers this directory as inputs of this task so that tests are triggered whenever there are updates to any of the dependent test resources.


stchar:
Thank you

PR #168, opened 2020-01-16
Proposal : a way to test small parts a big pipeline

kassovix:
I am using this kind of trick to test small parts of a big pipeline, or helper scripts that uses the pipeline execution context.
Tell me if you find this usage usefull (see TestRunClosure for examples).
I am not an expert of Groovy reflection, maybe there a simpler way to achieve this, please let me know.
Regards

PS : you have down a very good job with this library ;)

### DIFF
--- a/src/test/groovy/com/lesfurets/jenkins/TestRunClosure.groovy
+++ b/src/test/groovy/com/lesfurets/jenkins/TestRunClosure.groovy
@@ -20,7 +20,7 @@ class TestRunClosure extends BasePipelineTest {
     @Test
     void should_execute_without_errors() throws Exception {
         runScript({ script ->
-            script.echo 'Test'
+            echo 'Test'
         })
         printCallStack()
     }
@@ -28,7 +28,7 @@ class TestRunClosure extends BasePipelineTest {
     @Test
     void should_print_property_value() {
         runScript({ script ->
-            script.println 'value'
+            println 'value'
         })
 
         def value = 'value'
@@ -43,7 +43,7 @@ class TestRunClosure extends BasePipelineTest {
     void should_use_registered_method() {
         helper.registerAllowedMethod("customMethod", [Map.class], null)
         runScript({ script ->
-            script.customMethod test: 'value'
+            customMethod test: 'value'
         })
     }
 
@@ -58,8 +58,7 @@ class TestRunClosure extends BasePipelineTest {
                 .build()
         helper.registerSharedLibrary(library)
         runScript({ script ->
-
-            script.sayHello()
+            sayHello()
         })
     }
 }
--- a/src/test/groovy/com/lesfurets/jenkins/TestRunClosure.groovy
+++ b/src/test/groovy/com/lesfurets/jenkins/TestRunClosure.groovy
@@ -1,0 +1,47 @@
+package com.lesfurets.jenkins
+
+import com.lesfurets.jenkins.unit.BasePipelineTest
+import org.junit.Before
+import org.junit.Test
+
+import static com.lesfurets.jenkins.unit.MethodCall.callArgsToString
+import static org.junit.Assert.assertTrue
+
+class TestRunClosure extends BasePipelineTest {
+
+    @Override
+    @Before
+    void setUp() throws Exception {
+        super.setUp()
+    }
+
+    @Test
+    void should_execute_without_errors() throws Exception {
+        runScript({ script ->
+            script.echo 'Test'
+        })
+        printCallStack()
+    }
+
+    @Test
+    void should_print_property_value() {
+        runScript({ script ->
+            script.println 'value'
+        })
+
+        def value = 'value'
+        assertTrue(helper.callStack.findAll { call ->
+            call.methodName == 'println'
+        }.any { call ->
+            callArgsToString(call).contains(value)
+        })
+    }
+
+    @Test
+    void should_use_registered_method() {
+        helper.registerAllowedMethod("customMethod", [Map.class], null)
+        runScript({ script ->
+            script.customMethod test: 'value'
+        })
+    }
+}
--- a/src/test/groovy/com/lesfurets/jenkins/TestRunClosure.groovy
+++ b/src/test/groovy/com/lesfurets/jenkins/TestRunClosure.groovy
@@ -5,6 +5,8 @@ import org.junit.Before
 import org.junit.Test
 
 import static com.lesfurets.jenkins.unit.MethodCall.callArgsToString
+import static com.lesfurets.jenkins.unit.global.lib.LibraryConfiguration.library
+import static com.lesfurets.jenkins.unit.global.lib.ProjectSource.projectSource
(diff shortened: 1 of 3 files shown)


src/test/groovy/com/lesfurets/jenkins/TestRunClosure.groovy
```diff
@@ -0,0 +1,47 @@
...
+class TestRunClosure extends BasePipelineTest {
+
+    @Override
+    @Before
+    void setUp() throws Exception {
+        super.setUp()
+    }
+
+    @Test
+    void should_execute_without_errors() throws Exception {
```
stchar: Just an UX question: Do you want to test an exactly closure?
Or you would like to test an "inline" script 
```
runScript("""
   echo Hello World!
""")
```

src/test/groovy/com/lesfurets/jenkins/TestRunClosure.groovy
```diff
@@ -0,0 +1,47 @@
...
+        runScript({ script ->
```
stchar: I like the idea. Can I just use something like this?
```
runScript({
  echo "Hello World!"
})
```

src/test/groovy/com/lesfurets/jenkins/TestRunClosure.groovy
```diff
@@ -0,0 +1,47 @@
...
+        })
+    }
+
+    @Test
+    void should_use_registered_method() {
+        helper.registerAllowedMethod("customMethod", [Map.class], null)
+        runScript({ script ->
+            script.customMethod test: 'value'
+        })
+    }
```
stchar: could you please add a test case with a shared library?

src/test/groovy/com/lesfurets/jenkins/TestRunClosure.groovy
kassovix: It is exactly what I was trying to do in the first place, but I could not find a way to plug the framework on the closure (see [these 3 lines](https://github.com/jenkinsci/JenkinsPipelineUnit/pull/168/files#diff-517acc3404332fdd79708acaa681880eR191-R193)). It fails with this error:
`No signature of method: com.lesfurets.jenkins.TestRunClosure.echo()`

src/test/groovy/com/lesfurets/jenkins/TestRunClosure.groovy
kassovix: My use case is to test a part of code called in a big pipeline, that's why I did it with a closure.
But an inline script may also be interesting, I will give it a try!

src/test/groovy/com/lesfurets/jenkins/TestRunClosure.groovy
kassovix: I tried but I am not sure how to implement this. Could you take a look at my last commit?

stchar: Just tried locally your changes, I was unable to load library with `@Library` annotation, but successfully loaded it dynamically.

Good news classes are working fine, but methods from `vars` not exists.

I believe you have to dig into groovy interceptors    

```
@Test
    void should_use_library_sayHello() {
        String sharedLibs = this.class.getResource('/libs/commons@master').getFile()

        def library = library('commons')
                .defaultVersion('<notNeeded>')
                .allowOverride(true)
                .implicit(false)
                .targetPath('<notNeeded>')
                .retriever(projectSource(sharedLibs))
                .build()

        helper.registerAllowedMethod("library", [String.class], {String expression ->
            helper.getLibLoader().loadLibrary(expression)
            println helper.getLibLoader().libRecords
            return new LibClassLoader(helper,null)
        })

        runScript({ script ->
            def lib = script.library 'commons'
            def utils = lib.net.courtanet.jenkins.Utils.new(script)
            assert utils.gitTools() == [branch: 'master']
            script.sayHello()
        })
    }
```

stchar: I can only suggest to look at https://github.com/jenkinsci/JenkinsPipelineUnit/blob/master/src/main/groovy/com/lesfurets/jenkins/unit/PipelineTestHelper.groovy#L355
it calls private instance of GroovyScriptEngine which contains all the groovy magic with pipeline methods interception and loading of shared libraries.

the workaround might be to save your closure to a `String` and load it via `loadScript`

stchar:
This is weirdly but it works 
* Closure runner
```
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -1,6 +1,7 @@
 package com.lesfurets.jenkins.unit
 
 import static org.assertj.core.api.Assertions.assertThat
+import static groovy.lang.Closure.DELEGATE_ONLY
 
 abstract class BasePipelineTest {
 
@@ -181,17 +182,11 @@ abstract class BasePipelineTest {
      * @return the return value of the script
      */
     Object runScript(Closure scriptRun) {
-        final Script script = new Script(binding) {
-            @Override
-            Object run() {
-                scriptRun.call(this)
-                return null
-            }
-        }
-        script.metaClass.invokeMethod = helper.getMethodInterceptor()
-        script.metaClass.static.invokeMethod = helper.getMethodInterceptor()
-        script.metaClass.methodMissing = helper.getMethodMissingInterceptor()
-        return runScript(script)
+        File file = File.createTempFile("temp",".groovy")
+        def script = loadScript(file.absolutePath)
+        def rehydrate = scriptRun.rehydrate(script, script, script)
+        rehydrate.resolveStrategy = DELEGATE_ONLY
+        rehydrate.call()
     }

```

* test cases
```
void should_execute_without_errors() throws Exception {
         runScript({ script ->
-            script.echo 'Test'
+            echo 'Test'
         })
         printCallStack()
     }
 
```

```
-        runScript({ script ->
-
-            script.sayHello()
+        runScript({
+            sayHello()
         })

```

src/test/groovy/com/lesfurets/jenkins/TestRunClosure.groovy
stchar: no need to save closure in a file. We just need an instance of Script (the script could be empty) loaded with JPU's custom `GroovyScriptEngine`

so mocked pipeline methods would work, and library classes would be available. And then we just rehydrate the closure to this empty script. (Inspired with previously implemented support 
 of declarative pipeline ) #13

stchar:
hi @kassovix I'm going to release a new version of JPU next week, would you like to finish this feature? Do you need any help?

kassovix:
Thanks for your consideration but I won't have enough time to spend on this until next week.
You can proceed without this PR, and this will be for the following release ;)
Regards.

stchar:
@reinholdfuereder  yep this is definitely same as #382. I would close this one in favor of #382.

PR #174, opened 2020-02-14
Sort mocks in registerAllowedMethods

nre-ableton:
ping @stchar


stchar:
oh, my bad. will keep it sorted.

nre-ableton:
@stchar No problem, it's nobody's fault, but these things just happen over time. 😄

PR #175, opened 2020-02-14
Improve some mock implementations, add some new no-op mocks

nre-ableton:
This PR moves some mock implementations from the [`jenkins-pipeline-mocks`](https://github.com/AbletonDevTools/jenkins-pipeline-mocks) project to this one. Ping @stchar.

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -89,7 +89,9 @@ abstract class BasePipelineTest {
         helper.registerAllowedMethod("cron", [String.class], null)
         helper.registerAllowedMethod("dir", [String.class, Closure.class], null)
         helper.registerAllowedMethod("disableConcurrentBuilds", [], null)
-        helper.registerAllowedMethod("echo", [String.class], null)
+        helper.registerAllowedMethod("echo", [String.class]) { String message ->
+            println(message)
+        }
         helper.registerAllowedMethod("error", [String.class], { updateBuildStatus('FAILURE') })
         helper.registerAllowedMethod("gatlingArchive", [], null)
         helper.registerAllowedMethod("gitlabBuilds", [Map.class, Closure.class], null)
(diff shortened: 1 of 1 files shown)


nre-ableton:
Ah sorry, I forgot to run the test suite before pushing. 🤦‍♂ I'll push some fixups soon.

nre-ableton:
Ok, I see that the problem was with ada5282 (_Make the error mock throw an exception_). I think we need to have a larger discussion about this change in a separate PR. I've hence dropped it from this PR and force-pushed.

src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
```diff
@@ -100,24 +105,30 @@ abstract class BasePipelineTest {
...
         helper.registerAllowedMethod("logRotator", [Map.class], null)
+        helper.registerAllowedMethod('mail', [Map.class], null)
         helper.registerAllowedMethod("node", [Closure.class], null)
         helper.registerAllowedMethod("node", [String.class, Closure.class], null)
         helper.registerAllowedMethod("pipelineTriggers", [List.class], null)
         helper.registerAllowedMethod("properties", [List.class], null)
         helper.registerAllowedMethod("readFile", [String.class], null)
-        helper.registerAllowedMethod('retry', [Integer.class, Closure.class], null)
+        helper.registerAllowedMethod('retry', [Integer.class, Closure.class]) { Integer count, Closure body ->
+            body()
```
stchar: other mock uses [helper.callClosure](https://github.com/jenkinsci/JenkinsPipelineUnit/blob/85861a35547e4f6aea018790745ca6d52036b907/src/main/groovy/com/lesfurets/jenkins/unit/PipelineTestHelper.groovy#L499)
I'm not sure why, cannot see any unit test covering it, but I prefer to keep re-use same approach.
```
body.delegate = delegate
helper.callClosure(body)
```

nre-ableton: Great point, I've pushed 2 fixups to address this for `dir` and `retry`. Please let me know when you are OK with the state of this PR and I'll squash them before merge.

As for the use-case, here's an example: https://github.com/AbletonDevTools/postgres-pipeline-utils/blob/be944d643a0528a605c959b152426dd227bc10a5/src/com/ableton/PostgresDocker.groovy#L107-L114

We have several unit tests for the above code which assert that the stuff inside of `retry` ran correctly.

stchar: thank you

stchar:
@nre-ableton  thank you

nre-ableton:
@stchar Thanks!

Issue #152, opened 2019-08-29
Support inner enums


winterDroid:
Hello!

I have a script extracted which holds an `enum` like:
```groovy
// utils.groovy

enum Color {
    GREEN,
    RED
}

def printColor(Color color) {
    // ...
}

// To be visible in the pipeline
this.Color = Color

return this
```

In the actual pipeline I can access `Color` as follows:
```groovy
def utils = load "utils.groovy"
utils.printColor(utils.Color.RED)
```

When I now try to write a test for that with
```groovy
@Test
void "printColor should echo color"() {
    def script = loadScript("utils.groovy")
    script.printColor(script.Color.RED)
    testNonRegression()
}
```
it fails with the following exception:
```
groovy.lang.MissingPropertyException: No such property: Color for class: utils
	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.unwrap(ScriptBytecodeAdapter.java:66)
	at org.codehaus.groovy.runtime.callsite.PogoGetPropertySite.getProperty(PogoGetPropertySite.java:51)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callGetProperty(AbstractCallSite.java:299)
	at UtilsTest.printColor should echo color(UtilsTest.groovy:36)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:55)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
[...]

PR #187, opened 2020-03-13
Add tests to reproduce #152

stchar:
Reproduce #152


stchar:
It was a wrong usage of loadScript in #152

PR #191, opened 2020-03-26
Make closure to registerAllowedMethod optional

TobiX:
For your consideration - I'm fine if this is considered too dangerous 😺 - just close and forget the pull request then

Before, you had to specify the closure to registerAllowedMethod, even
if you wanted the default behaviour. For this reason, a big block of
registerAllowedMethod calls has a lot of noise, because each line ends
with `, null)`. This change makes it possible to omit.

Technically, this change is not API-compatible, but I think the benefits
outweigh the small amount of corner-cases this might break...

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -68,71 +68,71 @@ abstract class BasePipelineTest {
 
     void registerAllowedMethods() {
         // Please keep this list sorted when adding new mocks!
-        helper.registerAllowedMethod("addHtmlBadge", [Map.class], null)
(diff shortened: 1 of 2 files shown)


TobiX:
> I think that [5a0e631](https://github.com/jenkinsci/JenkinsPipelineUnit/commit/5a0e6312bba8efada3768471bfb7c2c94f8a5d84) should be dropped. Unless I'm missing something, the execution will not be the same.

At least the captured call tree from the test `com.lesfurets.jenkins.TestRegression` is the same as before. I cannot prove that both are doing the same, since I'm not exactly sure what the "default" execution flow is... I can revert if that is not enough evidence...

TobiX:
Dropped the commit for now, I'll try to find more evidence for my claims :D

stchar:
Nice! Merged.

Issue #104, opened 2018-05-23
JenkinsPipelineUnit vs accessing-steps at the base level


venosov:
Hi, there are 2 ways for accessing to steps in a shared library at the base level (src/): https://jenkins.io/doc/book/pipeline/shared-libraries/#accessing-steps

JenkinsPipelineUnit doesn't support the approach:

```
// src/org/foo/Zot.groovy
package org.foo;

def checkOutFrom(repo) {
  git url: "git@github.com:jenkinsci/${repo}"
}

return this
```

stchar:
I think it is not a bug as there is a test to validate such a case:
 https://github.com/jenkinsci/JenkinsPipelineUnit/blob/912fb118a1cd27feb59d7e12feee40888f7c1006/src/test/groovy/com/lesfurets/jenkins/TestSharedLibraryWithLocalSourceRetriever.groovy#L47

PR #193, opened 2020-03-28
filx(SharedLibs): Update GCL interceptors

stchar:
This PR is an attempt to continue the work stated in #111 as the author is not able to work on it.

Plan is:
1. Rebase previously started work
2. Comment out previous changes which intend to fix the issue
3. Add tests to reproduce the issue
4. Apply fix again if issue was reproduced, merge only tests if it wasn't


src/main/groovy/com/lesfurets/jenkins/unit/global/lib/LibraryLoader.groovy
```diff
@@ -111,6 +111,7 @@ class LibraryLoader {
                 }
             }
             record.definedGlobalVars = globalVars
+            println record.definedGlobalVars.collect({k,v->"$k::$v"})
```
stchar: delete debug output

src/test/groovy/com/lesfurets/jenkins/TestInterceptingGCL.groovy
```diff
@@ -0,0 +1,243 @@
...
+    String sharedLibOne = this.class
+        .getResource('/libs/test_cross_vars_usage')
+        .getFile()
+
+    String sharedLibTwo = this.class
+        .getResource('/libs/test_cross_class_usage')
+        .getFile()
+
+    @Test
+    void test_vars_interop_library_loaded_with_implicity() throws Exception {
```
stchar: Add comments for each tests what it does and why

stchar:
It verifies #104. You can create library class and call pipeline methods without passing it trough pipeline reference.

src/test/groovy/com/lesfurets/jenkins/TestInterceptingGCL.groovy
```diff
@@ -0,0 +1,243 @@
...
+        runScript('job/library/cross_class/test_dynamic.jenkins')
+        printCallStack()
+        assertCallStack().contains("""ClassA.methodA()""")
+        assertCallStack().contains("""ClassA.sh(echo 'ClassA: I'm field of A')""")
+        assertCallStack().contains("""ClassB.methodB()""")
+        assertCallStack().contains("""ClassB.sh(echo 'ClassB: I'm field of B')""")
+    }
+
+    @Test
+    void test_cross_lib_interop_no_implicity_annotation() throws Exception {
```
stchar, line 225: Cross lib tests are incorrect. 
scenarios should be like:
1. load libA, libB
2. Test pipeline can call vars from the both libs
3. Test libB can call LibA vars
4. Test case if both LibA and LibB have a vars/method with the same signature (who wins?) 
4.1 Verify how it works in real jenins   

Test the same (if possible) with classes (what if you call a static Method). 
Verify the behaviour is the same in real Jenkins instance

src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
```diff
@@ -24,19 +24,54 @@ class InterceptingGCL extends GroovyClassLoader {
...
         Class clazz = super.parseClass(codeSource, shouldCacheSource)
-        interceptClassMethods(clazz.metaClass, helper)
+        interceptClassMethods(clazz.metaClass, helper, binding)
         return clazz
     }
+
+    @Override
+    Class<?> loadClass(String name) throws ClassNotFoundException {
+        // Source from: groovy-all-2.4.6-sources.jar!/groovy/lang/GroovyClassLoader.java:710
+        println "Loading class: $name"
```
stchar: delete debug output

src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
```diff
@@ -140,7 +164,18 @@ abstract class BasePipelineTest {
                 updateBuildStatus('UNSTABLE')
             }
         })
+        helper.registerAllowedMethod("withCredentials", [Map, Closure])
         helper.registerAllowedMethod("withCredentials", [List, Closure], withCredentialsInterceptor)
+        helper.registerAllowedMethod('withEnv', [List, Closure], { List list, Closure c ->
+            list.each {
+                //def env = helper.get
+                def item = it.split('=')
+                assert item.size() == 2, "withEnv list does not look right: ${list.toString()}"
```
sams-gleb, line 175: it's possible to have additional '=' inside the item, e.g. when passing GRADLE_OPTS. Does it make sense to have this validation in the common library?

PR #196, opened 2020-03-30
Support not cloning method call arguments

brianeray:
v1.3 introduced a guard against call stack arguments mutating away from
the values they had at the time of the original method call. The guard
is to clone the argument instances at the time of the call.

In most use cases this ideal. But for arguments whose types are not
`Cloneable` the guard turns the arg values into `String` instances.
This is good fallback behavior but can break tests that do
fine-grained validations against nested structures, for example
`UnmodifiableMap` instances.

In those use cases the pre-v1.3 behavior of preserving the original
references may be desirable. The new
`PipelineTestHelper.cloneArgsOnMethodCallRegistration` property can
be set by the test to disable the cloning.


brianeray:
Our codebase makes heavy use of `final` and immutable maps and lists. So our argument references never mutate during the course of a unit test. v1.3 broke several dozen of our tests that looked at those types of arguments. I have patched them so as to now pass, but the string parsing makes the tests very hard to follow.

stchar:
looks cool

brianeray:
Thank you. I understand the reason for the original fix, and also realize that any shop validating args via string parsing would never need my current patch.

But it's going to save us some pain. We've got a lot of nested `UnmodifiableMap` args.

PR #197, opened 2020-03-30
feat(mock): Add mock for timestamps

### LINKED ISSUE #114: Mocking closure declared in library
Hello,
We are starting to use JenkinsPipelineUnit. We are stuck with the following groovy problem:

we have a library in src/be/company/Generic.groovy which does the following:
```groovy
package be.company;
def time(Closure cls) {
  timestamps {
    // do things before
    cls.call()
    // do things after
  }
}
```
`timestamps` is coming from this plugin: https://plugins.jenkins.io/timestamper
We are trying to mock the `timestamps` call using helper in our test like this:

```groovy
  @Override
  @Before
  void setUp() throws Exception {
      super.setUp()
      def library = library().name('jenkins-commons')
                       .defaultVersion("master")
                       .allowOverride(true)
                       .implicit(true)
                       .targetPath('build/libs')
                       .retriever(localSource('build/libs'))
                       .build()
      helper.registerAllowedMethod("timestamps", [Closure.class], null)
  }
```
But this mock doesn't seem to have the correct signature as we receive this error:

```
groovy.lang.MissingMethodException: No signature of method: be.company.Generic.timestamps() is applicable for argument types: (be.company.Generic$_time_closure1) values: [be.company.Generic$_time_closure1@3058b282]
```

What would be the best way to declare the correct signature to this mock ?

Using Gradle 4.8 - Groovy 2.4.12 - JenkinsPipelineUnit 1.1

stchar:
Fixes #114

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -130,6 +130,7 @@ abstract class BasePipelineTest {
         helper.registerAllowedMethod("string", [Map], stringInterceptor)
         helper.registerAllowedMethod('timeout', [Map])
         helper.registerAllowedMethod("timeout", [Map, Closure])
+        helper.registerAllowedMethod("timestamps", [Closure])
         helper.registerAllowedMethod("unstable", [String], { updateBuildStatus('UNSTABLE') })
         helper.registerAllowedMethod('unstash', [Map])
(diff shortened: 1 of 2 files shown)


nre-ableton:
While I'm not against this PR, I think we should be a bit cautious  am a bit skeptical about adding mocks in this library for pipeline steps provided by plugins. IMHO, this library should only cover pipeline steps which are part of core Jenkins, otherwise we risk bloating it up with all sorts of mocks.

stchar:
@nre-ableton, oh haven't thought about it. You right. I would avoid support signatures for plugins.

Issue #200, opened 2020-04-16
Update documentation: unable to resolve class com.lesfurets.jenkins.unit.declarative


seacephos:
The example for a declarative pipeline test in the readme is missing the .* after the package. 

Should be:
> import com.lesfurets.jenkins.unit.declarative.*

PR #201, opened 2020-04-17
Fix import in declarative pipeline example

nre-ableton:
Fixes https://github.com/jenkinsci/JenkinsPipelineUnit/issues/200, ping @stchar


stchar:
Thank you, merged

Issue #199, opened 2020-04-16
DeclarativePipelineTest withEnv mock should call closure only once


bogdanmuresan:
Currently the closure in the withEnv mock is called multiple times (same number as env params)

```
                c.delegate = binding
                c.call()
```

should be moved outside the loop that adds the env vars, like so 

```
        helper.registerAllowedMethod('withEnv', [List, Closure], { List list, Closure c ->
            list.each {
                //def env = helper.get
                def item = it.split('=')
                assert item.size() == 2, "withEnv list does not look right: ${list.toString()}"
                addEnvVar(item[0], item[1])
            }
            c.delegate = binding
            c.call()
        })
```

stchar:
@bogdanmuresan 
You are right. Would you like to submit a PR?

bogdanmuresan:
PR opened here: https://github.com/jenkinsci/JenkinsPipelineUnit/pull/205

PR #205, opened 2020-04-24
Issue #199 - fix withEnv closure call

bogdanmuresan:
Move closure call outside loop to prevent calling it multiple times.


stchar:
merged, thank you

PR #214, opened 2020-05-18
retry mock should actually retry

jsok:
Make `retry` work as expected.

Given a Jenkinsfile like:

```groovy
node {
  retry(3) {
    closureUnderTest()
  }
}
```

I could write a test like:

```groovy
    @Test
    void testEventuallySucceeds() throws Exception {
        binding.setVariable('succeedAfterAttempts', 2)
        helper.registerAllowedMethod('closureUnderTest', [], {
            if (succeedAfterAttempts == 0 ) {
                echo "success"
            } else {
                succeedAfterAttempts--
                error "failure"
            }
        })
        executeScript('retryExample.jenkins')
```

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -130,9 +130,20 @@ abstract class BasePipelineTest {
         helper.registerAllowedMethod("properties", [List])
         helper.registerAllowedMethod("pwd", [], { 'workspaceDirMocked' })
         helper.registerAllowedMethod("readFile", [String])
-        helper.registerAllowedMethod('retry', [Integer, Closure], { Integer count, Closure c ->
-            c.delegate = delegate
-            helper.callClosure(c)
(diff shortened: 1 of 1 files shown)


src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
```diff
@@ -130,9 +130,20 @@ abstract class BasePipelineTest {
         helper.registerAllowedMethod("properties", [List])
         helper.registerAllowedMethod("pwd", [], { 'workspaceDirMocked' })
         helper.registerAllowedMethod("readFile", [String])
-        helper.registerAllowedMethod('retry', [Integer, Closure], { Integer count, Closure c ->
-            c.delegate = delegate
-            helper.callClosure(c)
+        helper.registerAllowedMethod("retry", [Integer, Closure], { Integer count, Closure c ->
+            def attempts = 0
+            while (attempts != count) {
```
nre-ableton, line 135:
```suggestion
            while (attempts <= count) {
```

Paranoid, I know, but if someone called this mock with a negative value for `count`, then it would loop forever.

PR #220, opened 2020-05-25
Add when support for tag and buildingTag()

halkeye:
from https://www.jenkins.io/doc/book/pipeline/syntax/#built-in-conditions

Doesn't do the full comparator support, but supports glob which is the default comparator


halkeye:
awesome! <3 from the late night hacking

How do you do releases? just once in a while?

nre-ableton:
I'm not exactly sure what the release procedure or cadence is, sorry.

PR #236, opened 2020-06-21
Fix to parallel to return results

choosy:
Hi,

This pull request attempts to parallel method interceptor so that it returns a dictionary with the results of each closure call, just like the real parallel step does.

Let me know if there's anything else to be added here.


stchar:
Hi @choosy, glad to see your pull-request, could you please add some details about the use-case:
* What is the problem it's solving
* link an issue if any

nre-ableton:
The `parallel` step does indeed return a map with the keys, but I couldn't find this API documented anywhere.

stchar:
I'm closing the PR as it's need reworking but there is no response from the author. 
Please reopen once you fix failed tests

Issue #224, opened 2020-05-27
v1.4 lib class preload causes MissingMethodException with lib var step accepting lib class argument


brianeray:
This is probably best explained by first looking at PR #223. Note the test resources at the bottom of the PR. The `Monster1` lib class and `monster1` lib custom step are a common pattern in our pipeline code base.

Prior to 1.4 the JPU framework had no problem matching the lib step's `call` method signature with a caller passing an instance of the lib class to it. With all the logging from the PR in place and [the class preload logic commented out](https://github.com/jenkinsci/JenkinsPipelineUnit/blob/20d5a9613b3a616b953d26f3ccdbb169e390aa4a/src/main/groovy/com/lesfurets/jenkins/unit/global/lib/LibraryLoader.groovy#L115-L123) the call to the step will be successful, and my diagnostic logging looks something like this.

```
Loading shared library test_cross_class_usage with version master
DEBUG_ARGTYPE LibraryLoader.doLoadLibrary lib class preload disabled

DEBUG_ARGTYPE PTH.setGlobalVars monster1.call.nativeParameterTypes=[class org.test.Monster1]
DEBUG_ARGTYPE PTH.setGlobalVars nativeParameterTypes*.hashCode()=[1169712890]
DEBUG_ARGTYPE PTH.setGlobalVars cachedParamClasses=[class org.test.Monster1]
DEBUG_ARGTYPE PTH.setGlobalVars cachedParamClasses*.hashCode()=[1169712890]

DEBUG_ARGTYPE PTH.getAllowedMethodEntry argsClasses=[class org.test.Monster1]
DEBUG_ARGTYPE PTH.getAllowedMethodEntry argsClasses*.hashCode()=[1169712890]
DEBUG_ARGTYPE PTH.getAllowedMethodEntry casted to paramTypes=[class org.test.Monster1]
DEBUG_ARGTYPE PTH.getAllowedMethodEntry casted to paramTypes*.hashCode()=[1169712890]

DEBUG_ARGTYPE MethodSignature.equals thisClazz=class org.test.Monster1; thatClazz=class org.test.Monster1
DEBUG_ARGTYPE MethodSignature.equals thisClazz.canonicalName=org.test.Monster1; thatClazz.canonicalName=org.test.Monster1
DEBUG_ARGTYPE MethodSignature.equals thisClazz.hashCode()=1169712890; thatClazz.hashCode()=1169712890
[...]

PR #244, opened 2020-07-09
Issue #224 support lazy library class loading

brianeray:
v1.4 introduced the preload of shared library classes alongside the
existing preload of vars (ie custom steps).

In most use cases this is ideal. However in some use cases library
steps that use library classes in their call arguments are subject
to mistaken `MissingMethodException` errors when those classes are
preloaded. The method interceptor becomes confused because at some
point the same types are loaded more that once, and the argument
passed to the interceptor appears to be of a different type than
the type registered in the interceptor.

In those use cases the pre-v1.4 behavior of lazy loading the library
classes is helpful. The new `LibraryLoader.preloadLibraryClasses`
property can be set by the test to disable the class preload.

### DIFF
--- a/README.md
+++ b/README.md
@@ -581,6 +581,55 @@ Test class example:
     loadScript("job/library/exampleJob.jenkins")
     printCallStack()
 ```
+
+### Library global variables accepting library class instances as arguments: troubleshooting
+You might have a library defining global variables that implement custom steps
+accepting library class instances as arguments. For example consider the
+following library class and global variable.
+
+```groovy
+package org.test
+
+class Monster1 {
+    String moniker
+
+    Monster1(String m) {
+      moniker = m
+    }
+}
+```
+
+```groovy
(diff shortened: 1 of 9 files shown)


brianeray:
I made several attempts to patch library class preload without creating another feature flag but could never correct our test failures. But in its favor the new flag does not publicize itself prominently.

stchar:
hi @brianeray, PR looks good, could you please also update README.

brianeray:
Yes, should have that committed today or Monday at the latest. Thanks.

stchar:
Cool! Thank you.

brianeray:
Does that work? I also add a some doc for #196 in the third commit. I can remove that commit if you don't want it mixed in with this one.

brianeray:
No problem. Do have an idea of when 1.6 will be released? We'd love to start using it.

Issue #240, opened 2020-07-02
Bug: withEnv step values persist after closure (tested in declarative pipeline)


solvingj:
`withEnv` step values are only supposed to last for the duration of the closure that is passed to it.  There is currently a bug with JenkinsPipelineUnit handling of `withEnv` step where the values persist to all steps and stages within the rest of the pipeline.  This is a significant problem.  Here's a reproducible example:

Here's a pipeline, save it as `withEnvBugSpec_.jenkins`
```
pipeline {
    agent any
    stages {
        stage('withenvbug') {
            steps {
                withEnv(["SOMEVAR=SOMEVAL"]) {
                    echo "SOMEVAR inside closure = ${env.SOMEVAR}"
                }
                //SOMEVAR should be null after closure, it's not
                echo "SOMEVAR outside closure = ${env.SOMEVAR}"
            }
        }
    }
}
```

Here's a test of that pipeline. Save it as `withEnvBugSpec.groovy`
```
package test


import com.lesfurets.jenkins.unit.declarative.DeclarativePipelineTest
import org.junit.Before
import org.junit.jupiter.api.Test

class withEnvBugSpec extends DeclarativePipelineTest {
    @Override
    @Before
    void setUp() throws Exception {
        super.setUp()
        helper.registerAllowedMethod("echo", [String.class], { String s -> println(s) })
    };

    @Test
    void withEnvVariablesLinger() throws Exception {
        setUp()
        String stem = this.getClass().getCanonicalName().replace(".", "/") + "_"
        File jenkinsfile = new File("test/${stem}.jenkins")
        runScript(jenkinsfile.path)
        printCallStack()
    }
}
```

Here's the output of the test which shows the issue in the stdout 
```
Executing on agent [label:any]
SOMEVAR inside closure = SOMEVAL
SOMEVAR outside closure = SOMEVAL
   withEnvBugSpec_.run()
      withEnvBugSpec_.pipeline(groovy.lang.Closure)
         withEnvBugSpec_.echo(Executing on agent [label:any])
         withEnvBugSpec_.stage(withenvbug, groovy.lang.Closure)
            withEnvBugSpec_.withEnv([SOMEVAR=SOMEVAL], groovy.lang.Closure)
               withEnvBugSpec_.echo(SOMEVAR inside closure = SOMEVAL)
            withEnvBugSpec_.echo(SOMEVAR outside closure = SOMEVAL)
```

stchar:
[...]

PR #246, opened 2020-07-12
Issue #240 Avoid withEnv leaking environment out of scope

Alex-Vol-SV:
Used example pipeline and added fixed code block.

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -183,14 +183,26 @@ abstract class BasePipelineTest {
         helper.registerAllowedMethod("withCredentials", [Map, Closure])
         helper.registerAllowedMethod("withCredentials", [List, Closure], withCredentialsInterceptor)
         helper.registerAllowedMethod('withEnv', [List, Closure], { List list, Closure c ->
+            def stashedEnv = [:]
+            try {
+                stashedEnv.putAll(binding.getVariable('env') as Map)
+            } catch (MissingPropertyException e) {
+                // 'env' not set yet?
+            }
+
             list.each {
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
```diff
@@ -189,8 +189,16 @@ abstract class BasePipelineTest {
                 assert item.size() == 2, "withEnv list does not look right: ${list.toString()}"
                 addEnvVar(item[0], item[1])
             }
-            c.delegate = binding
-            c.call()
+            try {
+                c.delegate = binding
+                c.call()
```
stchar, line 194: save/restore env looks to be a natural approach, also it needs additional testcase

Alex-Vol-SV: What would you do in addition? The test case I added already confirms that the environment values are not present after the restore. Please suggest what is missing.

stchar: I mean to add a testcase to cover nested `withEnv`. Same as I described in #247. Instead of removing a field from `env` map after `c.call()` you just need to save/restore entire `env`.

Alex-Vol-SV: Is there any use case of the environment getting modified by the inner closure and restoring the entire env would be invalid? I am not sure Jenkins supports that but I will test to make sure it works as we think. Easy to add a full save, restore the env instead of the second iteration.

Alex-Vol-SV: Added full save/restore of env value

Issue #247, opened 2020-07-12
withCredentials method fails to pass tests when usernamePassword is provided


Alex-Vol-SV:
When using a pipeline with a basic usage of withCredentials the execution fails because the interceptor does not handle `usernamePassword` credentials correctly. The defined interceptors return the Map from the parameter list directly to the withCredentials interceptor which attempts to use it as if it was a string.

The solution I came up with was to replace the interceptors for string and usernamePassword to return a list of environment keys and the withCredentials to process the accordingly.

Here is my fix, I will follow up with a pull request
```

    def stringInterceptor = { m -> [m.variable] }

    def usernamePasswordInterceptor = { m -> [m.usernameVariable, m.passwordVariable] }

    def withCredentialsInterceptor = { list, closure ->
        list.forEach { creds ->
            creds.each {
                def var = it.toString()
                binding.setVariable(var, var)
            }
        }
        closure.delegate = delegate
        def res = closure.call()
        list.forEach { creds ->
            creds.each {
                def var = it.toString()
                binding.setVariable(var, null)
            }
        }
        return res
    }
```

My pipeline had this basic code structure that caused the failed test although Jenkins Pipeline is quite happy to run the same
```
    withCredentials([
            usernamePassword( credentialsId: 'my_cred_id', usernameVariable: 'user', passwordVariable: 'pass' ),
            string( credentialsId: 'docker_cred', variable: 'docker_pass' ),
            string( credentialsId: 'ssh_cred', variable: 'ssh_pass' )
        ])
    {
          ... code referencing creds here
    }
```

PR #248, opened 2020-07-12
Issue #247 Fix withCredentials when usernamePassword is used

Alex-Vol-SV:
Added different handling for usernamePassword and adjusted the
withCredentials interceptor to deal with it. The test job was created
and verified with the new code.


src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
```diff
@@ -25,13 +25,30 @@ abstract class BasePipelineTest {
...
+                binding.setVariable(creds, creds)                
+            } else {
+                creds.each {
+                    def var = it.toString()
+                    binding.setVariable(var, var)
+                }
+            }
         }
+        closure.delegate = delegate
         def res = closure.call()
```
stchar, line 42: Could you please use
```
helper.callClosure(closure)
```

src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
```diff
@@ -25,13 +25,30 @@ abstract class BasePipelineTest {
...
     def stringInterceptor = { m -> m.variable }
 
+    def usernamePasswordInterceptor = { m -> [m.usernameVariable, m.passwordVariable] }
+
     def withCredentialsInterceptor = { list, closure ->
-        list.forEach {
-            binding.setVariable(it, "$it")
+        list.forEach { creds ->
+            if (creds instanceof String) {
+                binding.setVariable(creds, creds)
```
stchar, line 33: I think it is a good idea to save previous value if any and then to restore it, 
im not sure if Jenkins does support nested calls but in your case
```
withCredentails(usernamePassword(['user', 'passwd'])) {
  sh 'somescript $user $passwd' // ok
  withCredentails(usernamePassword(['user', 'passwd'])) {
     sh 'another-script $user $passwd' // ok
  }
  sh 'somescript $user $passwd' // values would be null
}
```

src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
```diff
@@ -25,13 +25,30 @@ abstract class BasePipelineTest {
...
+
     def withCredentialsInterceptor = { list, closure ->
-        list.forEach {
-            binding.setVariable(it, "$it")
+        list.forEach { creds ->
+            if (creds instanceof String) {
+                binding.setVariable(creds, creds)                
+            } else {
+                creds.each {
+                    def var = it.toString()
```
stchar, line 36: You can remove `if (creds instanceof String) {` and keep just  `def var = it.toString()` 
java has `String.toString()` so groovy should have it too.

src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
stchar: could you please add a tests to verify credentails restore procedure

src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
Alex-Vol-SV: OK, I will do so. I used what was already there by the way.

src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
Alex-Vol-SV: I think this is probably an odd case but the original code did not do that. Since I am already in this area I might as well fix that. The same could be said for the withEnv code which I also changed to restore the environment by removing the added values.

src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
Alex-Vol-SV: That is not the reason for it. I tried to avoid regression in the call stacks. Existing call stacks of the previous `string` method as well as one of your unit tests reusing the stringInterceptor for `file` method would change if I used a common list return for both cases.

The returned values from the two interceptors are not the same. One returns a single value as string, the other returns a list of string which is why creds.each is involved. I need to know if the value is a string or a list. The it.toString is probably overly cautious handling of the value which is expected to be a String or GString for the credential environment key provided by usernameVariable and passwordVariable.

src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
Alex-Vol-SV: Done, now using callClosure

src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
Alex-Vol-SV: Done

stchar:
fix #247

Issue #239, opened 2020-07-01
Bug:  Unable to use "${param.SOMEVAR}" in an agent label


ballab1:
My pipeline script uses the following syntax:
`
pipeline {
    agent {
        label "branch-automation ${params.Location}"
    }
  ......
}
`
This causes a **MissingPropertyException**
`
groovy.lang.MissingPropertyException: No such property: params for class: com.lesfurets.jenkins.unit.declarative.AgentDeclaration
	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.unwrap(ScriptBytecodeAdapter.java:67)
	at org.codehaus.groovy.runtime.callsite.PogoGetPropertySite.getProperty(PogoGetPropertySite.java:51)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callGroovyObjectGetProperty(AbstractCallSite.java:310)
`

After looking through the issues, I thought this may be similar to previous issues with the **environment** Closure in a declarative script, but after looking through the code, I decided that there are significant differences. because of this, I cloned the  project, and made some changes to the _AgentDeclaration_ class, where I added a `Map<String,String> params = [:].` This did get me closer, as the exception I now get is 
`
lang.NullPointerException: Cannot get property 'Location' on null object
	at org.codehaus.groovy.runtime.NullObject.getProperty(NullObject.java:60)
	at org.codehaus.groovy.runtime.InvokerHelper.getProperty(InvokerHelper.java:190)
	at org.codehaus.groovy.runtime.callsite.NullCallSite.getProperty(NullCallSite.java:46)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callGetProperty(AbstractCallSite.java:299)
`

What I am now missing is:  how to populate this Map when reading a declarative pipeline script.

If you can give some hints, I'll get this working, update the unit tests and submit a PR

stchar:
Hi @ballab1, you right there is an issue #180 which is related to yours, and PR #186 with some tests to reproduce it.
Unfortunately I don't know how to fix it so far (( 
It is related to the closure delegate mechanics (IMHO).
Agent closure is executed without an access pipeline bindings.

ballab1:
I have provided a fix (for consideration)  - PR #252 
It includes a new Jenkins file to reproduce the issue.
[...]

PR #252, opened 2020-07-18
fix issue where agent closure with a label referencing a param variable cuases test failures

ballab1:
Issue #239:  Unable to use "${param.SOMEVAR}" in an agent label


ballab1:
realized there was a condition where the values of mocked parameters would get overwritten by content in the Jenkinsfile.
Fixed class so that if a param is defined it ill not be overwritten.This allows params to be mocked (before a test),and the mocked value to persist for the test

Issue #260, opened 2020-08-12
Global Var with single argument gets IllegalArgumentException if caller passed NULL


Alex-Vol-SV:
Having a Global Var with a single argument call will fail with IllegalArgumentException when the calling Jenkinsfile passes NULL in the argument list. This is not unusual and I traced the bug in the code that handles calling globals loaded from shared libraries.

I have a pull request prepared with the fix and will submit shortly.

This simple pipeline will cause the problem
```
@Library('commons@feature') _

// Test calling a one argument global with null
oneArg('Not null')
oneArg(null)
```

The oneArg.groovy global is just this simple method
```
// file: vars/oneArg.groovy
def call(arg) {
    if (arg == null) {
        echo "Is null"
    } else {
        echo "Is not null"
    }
}
```

stchar:
Nice catch!

PR #261, opened 2020-08-12
Issue #260: Fix IllegalArgumentException with GlobalVar taking a single argument

Alex-Vol-SV:
When calling a GlovalVar declared to accept one argument and the caller
passes null to the function it yields an IllegalArgumentException
instead of passing the null argument. This is unique to single argument
method calls and appears rooted in the way Grrovy lang handles the
invocation when null is provided.

This fix was inspired from the code in the invoked Groovy methods.

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/PipelineTestHelper.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/PipelineTestHelper.groovy
@@ -454,7 +454,16 @@ class PipelineTestHelper {
                 SCRIPT_SET_BINDING.invoke(script, binding)
                 script.metaClass.getMethods().findAll { it.name == 'call' }.forEach { m ->
                     this.registerAllowedMethod(method(e.value.class.name, m.getNativeParameterTypes()),
-                                    { args -> m.doMethodInvoke(e.value, args) })
+                        { args ->
+                            // When calling a one argument method with a null argument the
+                            // Groovy doMethodInvoke appears to incorrectly assume a zero
+                            // argument call signature for the method yielding an IllegalArgumentException
+                            if (args == null && m.getNativeParameterTypes().size() == 1) {
+                                m.doMethodInvoke(e.value, MetaClassHelper.ARRAY_WITH_NULL)
+                            } else {
+                                m.doMethodInvoke(e.value, args)
+                            }
+                        })
                 }
             }
             binding.setVariable(e.key, e.value)
--- a/src/test/resources/libs/commons@feature/vars/oneArg.groovy
+++ b/src/test/resources/libs/commons@feature/vars/oneArg.groovy
@@ -1,8 +1,4 @@
 // file: vars/oneArg.groovy
 def call(arg) {
-    if (arg == null) {
-        echo "Is null"
-    } else {
-        echo "Is not null"
-    }
+    echo arg ? "Is not null" : "Is null"
 }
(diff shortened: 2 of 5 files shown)


src/main/groovy/com/lesfurets/jenkins/unit/PipelineTestHelper.groovy
```diff
@@ -454,7 +454,16 @@ class PipelineTestHelper {
...
+                        { args ->
+                            // When calling a one argument method with a null argument the
+                            // Groovy doMethodInvoke appears to incorrectly assume a zero
+                            // argument call signature for the method yielding an IllegalArgumentException
+                            if (args == null && m.getNativeParameterTypes().size() == 1) {
+                                m.doMethodInvoke(e.value, MetaClassHelper.ARRAY_WITH_NULL)
+                            } else {
+                                m.doMethodInvoke(e.value, args)
+                            }
+                        })
```
nre-ableton, line 466:
```suggestion
                        { args -> m.doMethodInvoke(e.value, args ?: MetaClassHelper.ARRAY_WITH_NULL) })
```

The Elvis operator is a bit more succinct here.

src/test/resources/libs/commons@feature/vars/oneArg.groovy
```diff
@@ -0,0 +1,8 @@
+// file: vars/oneArg.groovy
+def call(arg) {
+    if (arg == null) {
+        echo "Is null"
+    } else {
+        echo "Is not null"
+    }
```
nre-ableton, line 7:
```suggestion
    echo arg ? "Is not null" : "Is null"
```

This can also be condensed into a one-liner.

src/main/groovy/com/lesfurets/jenkins/unit/PipelineTestHelper.groovy
Alex-Vol-SV: I do not believe it is for all use cases. The additional check for number of parameters is needed. args may be null for a zero argument method call.

nre-ableton: Ah, I see. Shouldn't it be `m.getNativeParameterTypes().size() > 1` then?

Alex-Vol-SV: In my opinion this code in Groovy is incorrectly handling the `null` case
[invoked Groovy](https://github.com/apache/groovy/blob/master/src/main/java/org/codehaus/groovy/reflection/ParameterTypes.java#L161)

The line in question should assign EMPTY_ARRAY to the null argumentArray and allow the subsequent lines decide if ARRAY_WITH_NULL is instead appropriate.

Alex-Vol-SV: No, more than one null arguments would have manifested as array of null. I can add additional tests for the behavior with multiple arguments but I believe the null single argument is an edge case not handled properly.

nre-ableton: Ok, I'm a bit unfamiliar with the background of the original failure here. I need to sit down and read through the linked issue before approving this PR. However, if another maintainer sees that everything is alright, I'm fine with them merging this (hence my leaving neutral feedback and not "requested changes").

PR #263, opened 2020-08-13
Fix arguments processed by mocked closures

nre-ableton:
These closures are called with two arguments: the arguments passed to
the step, and the closure to execute. If there is no placeholder
variable for the arguments, then this will result in the helper
mistakenly attempting to execute the arguments as a closure.


Alex-Vol-SV:
Great fix, I noticed the same and was thinking about preparing a PR of my own. You beat me to it. Good work!

PR #267, opened 2020-08-17
Handle default return correctly when returnStatus is used

Alex-Vol-SV:
The original default behavior for the `sh` step was undefined and fell
through the inaccurate handling of the resturnStdout behavior. This will
result in a success returnStatus when requested instead of a string
return.

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/PipelineTestHelper.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/PipelineTestHelper.groovy
@@ -683,7 +683,7 @@ class PipelineTestHelper {
         if (exitValue != 0) {
             throw new Exception('Script returned error code: ' + exitValue)
         }
-        return // no default return value requested
+        return null
     }
 
 }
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/lesfurets/jenkins/unit/PipelineTestHelper.groovy
```diff
@@ -680,6 +683,7 @@ class PipelineTestHelper {
         if (exitValue != 0) {
             throw new Exception('Script returned error code: ' + exitValue)
         }
+        return // no default return value requested
```
Alex-Vol-SV, line 686: IMHO this is a more accurate mock of the behavior of `sh` when no return is requested

Alex-Vol-SV:
For what it is worth, the original code also returns a multiline string as output when returnStdout=false which failed to pass when returnStatus=true because of the missing condition. Additionally, the default behavior of returnStdout=false and returnStatus=false is not implemented accurately in this block of code as it will still fall through to the multiline string return.

PR #268, opened 2020-08-22
Add support for "not" in "when" directive

IuryAlves:
Implementing `not` for `when` directive.

As docs says:

```
Execute the stage when the nested condition is false. Must contain one condition. For example: when { not { branch 'master' }}
```


stchar (approved):
Thank you

Issue #269, opened 2020-08-26
BasePipelineTest.runScript(Script script) throws MissingMethodException


Willem1987:
helper.runScript(script, this.binding) does not exists, should be helper.runScript(script)

groovy.lang.MissingMethodException: No signature of method: com.lesfurets.jenkins.unit.PipelineTestHelper.runScript() is applicable for argument types: (buildFromGitlab, groovy.lang.Binding) values: [buildFromGitlab@29852487, groovy.lang.Binding@771db12c]
Possible solutions: runScript(java.lang.String, groovy.lang.Binding), runScript(groovy.lang.Script), runScript(java.lang.String), loadScript(java.lang.String, groovy.lang.Binding)

	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.unwrap(ScriptBytecodeAdapter.java:58)
	at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.call(PogoMetaClassSite.java:54)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:48)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:113)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:133)
	at com.lesfurets.jenkins.unit.BasePipelineTest.runScript(BasePipelineTest.groovy:315)
	at org.codehaus.groovy.vmplugin.v7.IndyInterface.selectMethod(IndyInterface.java:232)
	at BuildTest.testBuild(BuildTest.groovy:33)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:132)
	at org.testng.internal.TestInvoker.invokeMethod(TestInvoker.java:599)
	at org.testng.internal.TestInvoker.invokeTestMethod(TestInvoker.java:174)
	at org.testng.internal.MethodRunner.runInSequence(MethodRunner.java:46)
[...]

PR #270, opened 2020-08-26
Correct helper.runscript call for loaded script

Willem1987:


nre-ableton (changes requested):
Please see my comment on the related issue: https://github.com/jenkinsci/JenkinsPipelineUnit/issues/269#issuecomment-680852368

nre-ableton (approved):
Sorry for the misunderstanding, and thanks for the fix!

PR #281, opened 2020-09-16
Allow kubernetes agent to be constructed with a map/object

Willem1987:
In our project we use a custom version of:
https://github.com/liejuntao001/jenkins-k8sagent-lib/blob/master/vars/k8sagent.groovy

Which basically returns a Map to construct a kubernetes agent with. This fix should allow constructing the kubernetes agent declaration based on a Map/Object


stchar (approved):

stchar:
Awesome! thank you.

PR #286, opened 2020-09-16
Add container allowed method

Willem1987:
For kubernetes agents the container method can be used to execute code in a container. Added allowed method

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -122,6 +122,10 @@ abstract class BasePipelineTest {
         helper.registerAllowedMethod("choice", [Map])
         helper.registerAllowedMethod('cifsPublisher', [Map], {true})
         helper.registerAllowedMethod('cleanWs')
+        helper.registerAllowedMethod("container", [String, Closure], { String path, Closure c ->
+            c.delegate = delegate
+            helper.callClosure(c)
+        })
         helper.registerAllowedMethod('copyArtifacts', [Map], {true})
         helper.registerAllowedMethod("cron", [String])
         helper.registerAllowedMethod('deleteDir')
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -122,6 +122,10 @@ abstract class BasePipelineTest {
         helper.registerAllowedMethod("choice", [Map])
         helper.registerAllowedMethod('cifsPublisher', [Map], {true})
         helper.registerAllowedMethod('cleanWs')
+        helper.registerAllowedMethod("container", [Map, Closure], { Map args, Closure c ->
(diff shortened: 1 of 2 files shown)


nre-ableton:
Please remove all indentation changes in this branch, as they make this PR harder to review. I agree that there is some inconsistency in the codebase, but we need a better approach to solving it. This is already on the radar, please see https://github.com/jenkinsci/JenkinsPipelineUnit/issues/237.

Willem1987:
Should be cleaner now

nre-ableton:
Looking better; can you please squash https://github.com/jenkinsci/JenkinsPipelineUnit/pull/286/commits/97aff47e3cdf18f43524fbfe85ac494510d490e9 to https://github.com/jenkinsci/JenkinsPipelineUnit/pull/286/commits/6aa088ee046f182c893a92fa52682942bc07b313, and also squash https://github.com/jenkinsci/JenkinsPipelineUnit/pull/286/commits/b955fafa35b0aad4f08e9b02ddea2c19de8a5f54 to https://github.com/jenkinsci/JenkinsPipelineUnit/pull/286/commits/90611428aa48e13973fc63d7728cad356edfd5ce? Thanks!

Willem1987:
Okay i cleaned it up to be 1 commit

src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
```diff
@@ -122,6 +122,10 @@ abstract class BasePipelineTest {
         helper.registerAllowedMethod("choice", [Map])
         helper.registerAllowedMethod('cifsPublisher', [Map], {true})
         helper.registerAllowedMethod('cleanWs')
+        helper.registerAllowedMethod("container", [String, Closure], { String path, Closure c ->
+            c.delegate = delegate
+            helper.callClosure(c)
+        })
```
nre-ableton, line 128:
```suggestion
        helper.registerAllowedMethod("container", [Map, Closure], { Map args, Closure c ->
            c.delegate = delegate
            helper.callClosure(c)
        })
        helper.registerAllowedMethod("container", [String, Closure], { String path, Closure c ->
            c.delegate = delegate
            helper.callClosure(c)
        })
```

I'm not terribly familiar with the Kubernetes plugin, but it seems that it supports taking both a `String` and a `Map`. It would be nice if this PR supported both.

Willem1987: A have only seen container used like this so far. Do you have an example? 

As far as i see here it just refers to a previously defined name:
https://github.com/jenkinsci/kubernetes-plugin

ContainerTemplate is a different story but i thought that was more of an agent like thing to run under.

Willem1987: https://www.jenkins.io/doc/pipeline/steps/kubernetes/ 
I see you are right. Couldn't find it at first. But it allows a name with shell combo.

nre-ableton (approved):
I'll wait a bit to see if anyone else wants to review this before merging. If nobody else reviews by Thursday (the 24th), I'll merge.

PR #294, opened 2020-09-29
Register timeout with integer argument

nre-ableton:
When an integer is given, it is assumed to be a value in minutes.


stchar (approved):

stchar:
Merged

src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
```diff
@@ -211,6 +211,10 @@ abstract class BasePipelineTest {
         helper.registerAllowedMethod("stage", [String, Closure])
         helper.registerAllowedMethod("step", [Map])
         helper.registerAllowedMethod("string", [Map], stringInterceptor)
+        helper.registerAllowedMethod('timeout', [Integer, Closure], { Map args, Closure c ->
```
gulyas-istvan-netlock-hu, line 214: Hello @nre-ableton !
`[Integer, Closure], { Map args, Closure c ->` is not type safe, better to use `[Integer, Closure], { Integer i, Closure c ->` here.

PR #299, opened 2020-10-01
Implement propertyMissing method

Willem1987:
Undo extending all Declarations from GenericPipelineDeclaration (this adds unwanted properties)
Remove GenericPipelineDeclaration implementation of getProperty.
Instead implement propertyMissing method to fallback to env and params.
Append test to use env var before pipeline closure
Remove duplicate method executeOn (use executeWith instead)
Remove name conflicts between JenkinsFile closures and Declaration fields

Our JenkinsFile had this:

```
def call() {
    String someValue = SomeHelper.someMap.get(ENV_VAR);

    pipeline {
    }
}
```

It seems the previous fix to handle looking into env. and params. does not work outside the pipeline closure. Therefore i looked for another way to fix this. 
During this i noticed the effect extending GenericPipelineDeclaration has on the function printNonNullProperties of ObjectUtils.
So i undid the extending of PipelineDeclaration for declarations that should not be able to handle all those functions(agent, steps etc).

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/declarative/AgentDeclaration.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/declarative/AgentDeclaration.groovy
@@ -4,10 +4,12 @@ import com.lesfurets.jenkins.unit.declarative.agent.DockerAgentDeclaration
 import com.lesfurets.jenkins.unit.declarative.agent.KubernetesAgentDeclaration
 import groovy.transform.ToString
 
+import static com.lesfurets.jenkins.unit.declarative.GenericPipelineDeclaration.createComponent
+import static com.lesfurets.jenkins.unit.declarative.GenericPipelineDeclaration.executeWith
 import static groovy.lang.Closure.DELEGATE_FIRST
 
 @ToString(includePackage = false, includeNames = true, ignoreNulls = true)
-class AgentDeclaration extends GenericPipelineDeclaration {
+class AgentDeclaration {
 
     String label
     DockerAgentDeclaration docker
@@ -16,7 +18,6 @@ class AgentDeclaration extends GenericPipelineDeclaration {
     String dockerfileDir
     Boolean reuseNode = null
     String customWorkspace
-    def binding = null
 
     def label(String label) {
         this.label = label
@@ -35,7 +36,7 @@ class AgentDeclaration extends GenericPipelineDeclaration {
     }
 
     def docker(String image) {
-        this.docker = new DockerAgentDeclaration().with { it.image = image; it }
+        this.docker({ -> this.image = image })
     }
 
     def docker(@DelegatesTo(strategy = DELEGATE_FIRST, value = DockerAgentDeclaration) Closure closure) {
@@ -60,18 +61,6 @@ class AgentDeclaration extends GenericPipelineDeclaration {
 
     def dir(String dir) {
         this.dockerfileDir = dir
-    }
-
-    def getCurrentBuild() {
-        return binding?.currentBuild
-    }
-
-    def getEnv() {
-        return binding?.env
-    }
-
-    def getParams() {
-        return binding?.params
     }
 
     def execute(Object delegate) {
--- a/src/main/groovy/com/lesfurets/jenkins/unit/declarative/AgentDeclaration.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/declarative/AgentDeclaration.groovy
@@ -4,9 +4,7 @@ import com.lesfurets.jenkins.unit.declarative.agent.DockerAgentDeclaration
 import com.lesfurets.jenkins.unit.declarative.agent.KubernetesAgentDeclaration
 import groovy.transform.ToString
 
-import static com.lesfurets.jenkins.unit.declarative.GenericPipelineDeclaration.createComponent
 import static com.lesfurets.jenkins.unit.declarative.GenericPipelineDeclaration.executeWith
-import static groovy.lang.Closure.DELEGATE_FIRST
 
 @ToString(includePackage = false, includeNames = true, ignoreNulls = true)
 class AgentDeclaration {
@@ -23,7 +21,7 @@ class AgentDeclaration {
         this.label = label
     }
 
-    def node(@DelegatesTo(AgentDeclaration) Closure closure) {
+    def node(Closure closure) {
         closure.call()
     }
 
@@ -36,26 +34,29 @@ class AgentDeclaration {
     }
 
     def docker(String image) {
-        this.docker({ -> this.image = image })
+        this.docker = new DockerAgentDeclaration().with{  da -> da.image = image; da }
     }
 
-    def docker(@DelegatesTo(strategy = DELEGATE_FIRST, value = DockerAgentDeclaration) Closure closure) {
-        this.docker = createComponent(DockerAgentDeclaration, closure)
+    def docker(Closure closure) {
+        this.docker = new DockerAgentDeclaration();
+        executeWith(this.docker, closure);
     }
 
     def kubernetes(Object kubernetesAgent) {
         this.@kubernetes = kubernetesAgent as KubernetesAgentDeclaration
     }
 
-    def kubernetes(@DelegatesTo(strategy = DELEGATE_FIRST, value = KubernetesAgentDeclaration) Closure closure) {
-        this.@kubernetes = createComponent(KubernetesAgentDeclaration, closure)
+    def kubernetes(Closure closure) {
+        this.@kubernetes = new KubernetesAgentDeclaration();
+        def kubernetesDecl = this.@kubernetes
+        executeWith(kubernetesDecl, closure, Closure.DELEGATE_FIRST)
     }
 
     def dockerfile(boolean dockerfile) {
         this.dockerfile = dockerfile
     }
 
-    def dockerfile(@DelegatesTo(AgentDeclaration) Closure closure) {
+    def dockerfile(Closure closure) {
         closure.call()
     }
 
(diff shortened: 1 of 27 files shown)


stchar (commented):
Have a question regarding reverting some changes

src/main/groovy/com/lesfurets/jenkins/unit/declarative/AgentDeclaration.groovy
```diff
@@ -4,10 +4,12 @@ import com.lesfurets.jenkins.unit.declarative.agent.DockerAgentDeclaration
 import com.lesfurets.jenkins.unit.declarative.agent.KubernetesAgentDeclaration
 import groovy.transform.ToString
 
+import static com.lesfurets.jenkins.unit.declarative.GenericPipelineDeclaration.createComponent
+import static com.lesfurets.jenkins.unit.declarative.GenericPipelineDeclaration.executeWith
 import static groovy.lang.Closure.DELEGATE_FIRST
 
 @ToString(includePackage = false, includeNames = true, ignoreNulls = true)
-class AgentDeclaration extends GenericPipelineDeclaration {
```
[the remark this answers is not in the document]
stchar: I've read the PR description the question is closed.

src/main/groovy/com/lesfurets/jenkins/unit/declarative/GenericPipelineDeclaration.groovy
```diff
@@ -14,24 +15,14 @@ abstract class GenericPipelineDeclaration {
     static <T> T createComponent(Class<T> componentType, @DelegatesTo(strategy = DELEGATE_FIRST) Closure<T> closure) {
         // declare componentInstance as final to prevent any multithreaded issues, since it is used inside closure
         final def componentInstance = componentType.newInstance()
```
stchar, line 17: I think at the moment we create a new instance of whatever we should call our [custom class loader](https://github.com/jenkinsci/JenkinsPipelineUnit/pull/299/files#diff-d0f7b520f331068851ec6f302d72e7998d2c75a27e56c924a224f47b44918be1).
the problem is `createComponent` method is static so I'm not sure how to retrieve it.

stchar: so no additional juggling with bindings are required in `execute`

src/main/groovy/com/lesfurets/jenkins/unit/declarative/PostDeclaration.groovy
```diff
@@ -59,48 +59,48 @@ class PostDeclaration {
...
             if((currentBuild == 'FAILURE' || currentBuild == 'UNSTABLE') && previousBuild == 'SUCCESS'){
-                executeOn(delegate, this.regression)
+                executeWith(delegate, this.regression)
             }
         }
 
         // Cleanup is always performed last
         if(this.cleanup){
-            executeOn(delegate, this.cleanup)
+            executeWith(delegate, this.cleanup)
```
stchar, line 103: `ExecuteOn` and `ExecuteWith` are the same so currently I dont see a point to rename them
 
I'm not sure what was the original idea to keep this methods separated maybe @ozangunalp  could add some comments as he was an initial author of declarative feature

src/main/groovy/com/lesfurets/jenkins/unit/declarative/StageDeclaration.groovy
```diff
@@ -1,31 +1,53 @@
 package com.lesfurets.jenkins.unit.declarative
 
 
+import static com.lesfurets.jenkins.unit.declarative.GenericPipelineDeclaration.createComponent
+import static com.lesfurets.jenkins.unit.declarative.GenericPipelineDeclaration.executeWith
 import static groovy.lang.Closure.*
 
-class StageDeclaration extends GenericPipelineDeclaration {
+class StageDeclaration {
```
stchar, line 8: stage declaration is actually need GenericPipleineDecalration IMHO

src/main/groovy/com/lesfurets/jenkins/unit/declarative/StageDeclaration.groovy
```diff
@@ -41,24 +63,42 @@ class StageDeclaration extends GenericPipelineDeclaration {
...
 
-        if(delegate.binding.variables.currentBuild.result == "FAILURE"){
+        if (delegate.currentBuild.result == "FAILURE") {
             executeWith(delegate, { echo "Stage \"$name\" skipped due to earlier failure(s)" })
             return
         }
 
         if (!when || when.execute(delegate)) {
-            super.execute(delegate)
+            // Set environment for stage
```
stchar, line 82: I would prefer not to copy-paste same code in different classes inheritance here is a good practice

src/test/jenkins/jenkinsfiles/Environment_Jenkinsfile
```diff
@@ -1,18 +1,24 @@
...
-    stages {
-        stage('Example') {
-            environment {
-                LEVAR1 = "LE NEW VALUE"
-                LEVAR2 = "A COPY OF ${env.LEVAR1} in build#${currentBuild.number}"
-            }
-            steps {
-                echo "LEVAR1 ${env.LEVAR1}"
-                echo "LEVAR2 ${env.LEVAR2}"
+def call() {
```
stchar, line 1: could you please add a comment why to wrap pipeline into a call method as it is not so obvious
If it is an additional testcase then create a new Jenkinsfile

src/main/groovy/com/lesfurets/jenkins/unit/declarative/StageDeclaration.groovy
Willem1987: Does StageDeclaration have all the things a PipelineDeclaration has? I was in doubt about this one...

src/main/groovy/com/lesfurets/jenkins/unit/declarative/PostDeclaration.groovy
Willem1987: I removed ExecuteOn, so i had to replace references. We could forward one to the other if we need it for backwards compatibility

src/main/groovy/com/lesfurets/jenkins/unit/declarative/StageDeclaration.groovy
Willem1987: The stage should clear the environment/params after finishing so the next stage could inherit pipeline vars. But i guess we could implement the cleanup at GenericPipeline level

src/test/jenkins/jenkinsfiles/Environment_Jenkinsfile
Willem1987: We wrap the entire pipeline script in a method in our project to be able to access it from a jenkins pipelinescript job.

In the jenkins job we don't have to have the scm url in each build. We just have 1 library which we can use var groovyscripts from.

Inspiration taken from:
https://www.jenkins.io/doc/book/pipeline/shared-libraries/

So basically we have a method we can call to build a pipeline block. We do some initialisation in the block above pipeline{

This way we can use custom labels depending on env or input params. This allows us to reuse pipelines over a multitude of jobs.

Willem1987: The call() at the bottom isn't required in jenkins, but i wanted to avoid going loadScript("scriptname")()

src/main/groovy/com/lesfurets/jenkins/unit/declarative/StageDeclaration.groovy
Willem1987: I went back to extending GenericPipelineDeclaration

src/main/groovy/com/lesfurets/jenkins/unit/declarative/StageDeclaration.groovy
Willem1987: I created a public static method to init and reset the environment props. This way we can share the logic, but choose when to call the parts.

We want to init the env in a stage, but only reset it after the stage is complete. Super.execute would have immediately cleared the stage custom env overrides.

src/main/groovy/com/lesfurets/jenkins/unit/declarative/GenericPipelineDeclaration.groovy
Willem1987: I'm sorry i don't quite follow. Are you saying we want to get rid of the createComponent call alltogether? The only change i did in this MR is flip 2 vars around. Making closure the delegate and componentInstance the owner. So that the closure is evaluated before accessing the Declaration class methods/vars

src/test/jenkins/jenkinsfiles/Environment_Jenkinsfile
Willem1987: I re-read the request and made a new testcase to show the scoping works.

src/main/groovy/com/lesfurets/jenkins/unit/declarative/GenericPipelineDeclaration.groovy
Willem1987: @stchar  I tried simplifying the whole class loading.

The idea:
We only need stack interception for Script calls? (not the declaration setup calls, as far as i understand it)
The declaration classes get called to set properties as a sort of setup of the stack. When execute is called on the Declaration it uses the Script to get the correct depth and stack logging to work.
The parameters declaration now uses the binding of the closure (coming from PipelineTestHelper/BasePipelineTest)

A problem i ran into when trying to intercept the declaration classes is calling super methods. More specifically the super.execute call in NotDeclaration. The current MethodInterceptor seems unable to distinguish when to call NotDeclaration.execute vs super(/WhenDeclaration).execute. It seems to keep trying to intercept and invoke NotDeclaration.execute. This caused when_not_branch_master to fail.

So from that i tried just making sure the closures get bound to a Declaration delegate and the Declaration class delegates calls throug the script to get through the Stack interception

PR #306, opened 2020-10-06
Add fileExists mock

cwiggs:
This commit adds the fileExist mock that returns "true".  Adding it here so people don't have to mock such a simple step when using the JenkinsPipelineUnit framework.

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -157,7 +157,7 @@ abstract class BasePipelineTest {
             println(message)
         })
         helper.registerAllowedMethod("error", [String], { updateBuildStatus('FAILURE') })
-        helper.registerAllowedMethod('fileExists', [String], {true}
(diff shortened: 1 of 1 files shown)


nre-ableton:
I don't think we should add a mock for this. Returning `true` in all cases makes a pretty broad assumption that isn't always true for pipelines in many cases. Leaving this step unmocked means that users must add their own mock for it, which I think is good to pass the responsibility to them.

If we were to add a mock for `fileExists` to this library, I'd prefer it to be similar to the `readFile` mock, meaning that there would be a [separate helper method](https://github.com/jenkinsci/JenkinsPipelineUnit/blob/master/src/main/groovy/com/lesfurets/jenkins/unit/PipelineTestHelper.groovy#L571) to register the output that should be returned for a given filename.

stchar (approved):

stchar (changes requested):
@nre-ableton is right additional logic is required to filter what files should exist in which tests.

nre-ableton:
See https://github.com/jenkinsci/JenkinsPipelineUnit/pull/312.

cwiggs:
#312 Looks good to me, thanks!  You can close this PR.

PR #311, opened 2020-10-16
fix(declarative) withCredentails usernamePassword mocking conflict

### LINKED ISSUE #310: withCredentials doesn't work with declarative pipeline after 1.8
```
groovy.lang.MissingMethodException: No signature of method: groovy.lang.Binding.getVariable() is applicable for argument types: (java.util.LinkedHashMap$Entry) values: [credentialsId=credid]
Possible solutions: getVariable(java.lang.String), getVariables(), setVariable(java.lang.String, java.lang.Object), hasVariable(java.lang.String)
	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.unwrap(ScriptBytecodeAdapter.java:58)
	at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.call(PogoMetaClassSite.java:54)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:48)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:113)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:125)
	at com.lesfurets.jenkins.unit.BasePipelineTest$_closure3$_closure31$_closure33.doCall(BasePipelineTest.groovy:60)
```

Declarative pipeline to reproduce 
```
pipeline {
   agent any
   stages {
      stage('build') {
         withCredentials([usernamePassword(credentialsId: 'credid', passwordVariable: 'pswd', usernameVariable: 'usr')]) {
           sh "run_some $pswd $usr"
         }
      }
   }
}

```

stchar:
Fixes #310

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/declarative/DeclarativePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/declarative/DeclarativePipelineTest.groovy
@@ -16,6 +16,19 @@ abstract class DeclarativePipelineTest extends BasePipelineTest {
         addParam(desc.name, desc.defaultValue, false)
     }
 
+    def stringInterceptor = { Map desc->
+        if (desc) {
+            // we are in context of paremetes { string(...)}
+            if (desc.name) {
+                addParam(desc.name, desc.defaultValue, false)
+            }
+            // we are in context of withCredentaials([string()..]) { }
+            if(desc.variable) {
+                return desc.variable
+            }
+        }
+    }
+
     @Override
     void setUp() throws Exception {
         super.setUp()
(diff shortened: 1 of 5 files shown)


nre-ableton (approved):
Nice fix!

Willem1987:
Question:

Why would we add a param in this case? The ParametersDeclaration has the string method when calling within a parameters closure.

When the string is called on the script outside the closure, i would not expect params to be added. That's why i did not go through the interceptor. Though i'm not sure if thats building on from #299 

I see a clear difference between the string method being called as a param to a downstream build vs the string method inside the parametersdeclaration. I feel only the parametersdeclaration should actually add params.

stchar:
@Willem1987 I cannot determine whether the string mock is called in which context (`parameters {}` or `withCredentials([...]) {...}` ) that's why I've implemented it as is

Willem1987:
@stchar If i'm correct the call within the Parameters block should end up in ParametersDeclaration class first before getting to the outer lying script. But i could be mistaken. The tests seemed to work the way i tried to implement it

stchar:
@Willem1987 
It looks like parameters block should have it's own `string` interceptor or to be updated to support interceptor for `withCredentials`:

registerAllowedMethod('string', [Map]) acts differently in DeclarativePipelineTest and BasePipelineTest 
But we hadn't seen any differences until `withCredentials` has been refactored in BasePipelineTest. 

DeclarativePipelineTest inherits BasePipelineTest so withCredentials interceptor had been inherited 
in DeclaratievePipelineTest which fired errors in my tests once I updated version of JPU up to 1.8

I would prefer ParametersDeclaration to shadow pipeline dsl but not to mock any pipeline plugins callbacks

PR #314, opened 2020-10-26
Allow use of ant path expressions on branch validation

epsilo:
Allow use of ant path expressions on branch validation


epsilo:
wrong

Issue #209, opened 2020-05-04
Need to manually add assertj-core to build.gradle dependencies


nre-ableton:
I'm in the process of updating some of my projects from JenkinsPipelineUnit 1.0 to 1.4, and I now need to add the following to my `build.gradle` to make it work:

```gradle
dependencies {
  compile 'com.github.AbletonDevTools:jenkins-pipeline-mocks:0.8+'
  compile 'org.codehaus.groovy:groovy-all:2.4.12'

  testCompile 'org.assertj:assertj-core:3.15+' // <-- this
  testCompile 'junit:junit:4.13'
  testCompile(group: 'com.lesfurets', name: 'jenkins-pipeline-unit', version: '1.4')
}
```

If I omit the `testCompile 'org.assertj:assertj-core:3.15+'` line, I get this error when running `./gradlew test`:

> `Task :compileTestGroovy FAILED`
> `FAILURE: Build failed with an exception.`
> `* What went wrong:`
> `Execution failed for task ':compileTestGroovy'.`
> `Unable to load class com.lesfurets.jenkins.unit.BasePipelineTest due to missing dependency org/assertj/core/api/AbstractCharSequenceAssert`

I'm not terribly skilled with Gradle, but I had a look at this project's `build.gradle` file and everything seemed to be in order. Am I missing something obvious?

TobiX:
Probably caused by #185, this wasn't the case with 1.3

brianeray:
I just hit this error in a slightly different context, as we have another Spock-specific framework sitting on top of JPU:

`
  implementation 'com.lesfurets:jenkins-pipeline-unit:1.4'
`

So the exception is thrown earlier in this project, during `:compileGroovy`.

I believe there's some difference between `compile` and `implementation`, and `testCompile` and `testImplementation` as in the #185 refactor. We have other projects in which switching from the former to the latter also caused transitive dependencies falling off the classpath.

brianeray:
The POMs in the 1.3 and 1.4 JARs show different dependency scopes.

```
@@ -78,5 +78,5 @@
       <artifactId>assertj-core</artifactId>
       <version>3.4.1</version>
-      <scope>compile</scope>
+      <scope>runtime</scope>
     </dependency>
```

That might be a downstream side effect of Gradle's deprecation of `compile` and `testCompile`.

nre-ableton:
[...]

PR #315, opened 2020-10-26
Allow use of ant path expressions on branch validation

epsilo:
Allow use of ant path expressions on branch validation

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,6 +37,7 @@ dependencies {
     implementation('org.apache.ivy:ivy:2.4.0')
     api('org.assertj:assertj-core:3.4.1')
     implementation('org.apache.commons:commons-lang3:3.5')
+    implementation('org.springframework:spring-core:5.2.9.RELEASE')
 
     testImplementation('junit:junit:4.11')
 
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -37,6 +37,7 @@ dependencies {
     implementation('org.apache.ivy:ivy:2.4.0')
     api('org.assertj:assertj-core:3.4.1')
     implementation('org.apache.commons:commons-lang3:3.5')
+    implementation('org.springframework:spring-core:5.2.9.RELEASE')
```
stchar, line 40: Looks like `api` should  be used instead of `implementation` if this library is needed in run-time. See https://github.com/jenkinsci/JenkinsPipelineUnit/issues/209

build.gradle
```diff
@@ -89,6 +90,16 @@ jacocoTestReport {
     }
 }
 
+task buildJarWithDependants(type: Jar) {
```
stchar, line 93: I suppose this is to workaround #209 so not really needed

stchar (changes requested):
Looks good to me, minor changes in gradle configuration are required.

stchar (approved):
gracias!

stchar:
merged, thank you for the efforts.

PR #321, opened 2020-11-23
set empty map for 'params'  in binding

noiseoverip:
without this. calling addParam() fails

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -291,7 +291,9 @@ abstract class BasePipelineTest {
         ])
         binding.setVariable('docker', new DockerMock())
         binding.setVariable('env', [:])
+        binding.setVariable('params', [:])
         binding.setVariable('scm', [:])
+        
     }
 
     /**
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -293,7 +293,6 @@ abstract class BasePipelineTest {
         binding.setVariable('env', [:])
         binding.setVariable('params', [:])
         binding.setVariable('scm', [:])
-        
     }
 
     /**
--- a/src/test/groovy/com/lesfurets/jenkins/TestParamInit.groovy
+++ b/src/test/groovy/com/lesfurets/jenkins/TestParamInit.groovy
@@ -1,0 +1,14 @@
+package com.lesfurets.jenkins
+
+import com.lesfurets.jenkins.unit.BasePipelineTest
+import org.junit.Test
+
+class TestParamInit extends BasePipelineTest {
+
+    @Test
+    void readUndefinedParamNoException() throws Exception {
+        scriptRoots += 'src/test/jenkins'
+        super.setUp()
+        runScript('job/printParams.jenkins')
+    }
+}
--- a/src/test/jenkins/job/printParams.jenkins
+++ b/src/test/jenkins/job/printParams.jenkins
@@ -1,0 +1,1 @@
+print params.FOO


src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
```diff
@@ -291,7 +291,9 @@ abstract class BasePipelineTest {
         ])
         binding.setVariable('docker', new DockerMock())
         binding.setVariable('env', [:])
+        binding.setVariable('params', [:])
         binding.setVariable('scm', [:])
+
```
nre-ableton, line 296:
```suggestion
```

nre-ableton (commented):
I need to play around with this branch a bit, as I'm unfamiliar with the `params` singleton, but I at least added a suggestion to fix a whitespace error.

nre-ableton:
I'm not convinced that this change is needed, though it is fairly innocuous. `BasePipelineTest.addParam()` can handle the case where `params` has not yet been added to the binding, see: https://github.com/jenkinsci/JenkinsPipelineUnit/blob/350a6e51fd58366c4aa302b6cbcc8782b50c5225/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy#L435-L444

Also, there is a test Jenkinsfile for this feature: `src/test/jenkins/jenkinsfiles/Params_Jenkinsfile`

Can you please show an example of a Jenkinsfile which fails without this patch?

stchar:
@noiseoverip, could you please add a test to reproduce the issue (for scritped and declarative pipelines)

noiseoverip:
Hi,
`BasePipelineTest.addParam()` doesn't help because it will throw exception when `params` is not initialized due to behavior of `getVariable()`. 

```    
public Object getVariable(String name) {
        if (variables == null)
            throw new MissingPropertyException(name, this.getClass());

        Object result = variables.get(name);

        if (result == null && !variables.containsKey(name)) {
            throw new MissingPropertyException(name, this.getClass());
        }

        return result;
    }
```

I think it is reasonable to initialize it by default since the same is done to `env` var and that would more closely match Jenkins behavior since script can call `params.FOO` under any circumstance (parameterized, non-parameterized) job and that doesn't throw any exceptions just returns NULL when param doesn't have any value. And yes i can create unit tests, just need to find time for this

nre-ableton:
Perhaps I am missing something obvious here, but I added a test based on your suggestion, in `TestRegression.groovy`:

```groovy
    @Test
    void testParamsMapInitialized() throws Exception {
        def script = runScript("job/params.jenkins")
        script.execute()
        super.testNonRegression("params")
    }
```

...with the corresponding Jenkinsfile:

```groovy
def execute() {
    echo params.getClass().toString()
    echo(params.FOO)
}
return this
```

The test fails, even if this patch is applied (condensed stacktrace):

```
class java.lang.Class

groovy.lang.MissingPropertyException: No such property: FOO for class: params

	at groovy.lang.MetaClassImpl.invokeStaticMissingProperty(MetaClassImpl.java:1001)
        ...
	at ___cps.transform___(Native Method)
	at com.lesfurets.jenkins.unit.cps.PipelineTestHelperCPS.callMethod(PipelineTestHelperCPS.groovy:32)
        ...
	at com.lesfurets.jenkins.TestRegression.testParamsMapInitialized(TestRegression.groovy:33)
        ...
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:53)
```

Note that the echo statement thinks that `params` is a `Class` object, not a `Map`.

noiseoverip:
I believe it fails because you don't have:
```    
@Override
@Before
    void setUp() throws Exception {
        super.setUp()
}
```

Since all the helper stuff doesn't get initialized if you don't do this

noiseoverip:
the stack trace you printed is from a run without a patch. Adding that patch fixes it though of course the test it self fails due to callstack now being different. I have pushed a simple unit test to prove my point. That unit test fails without the patch

JorisBolsens:
> Perhaps I am missing something obvious here, but I added a test based on your suggestion, in `TestRegression.groovy`:
> 
> ```groovy
>     @Test
>     void testParamsMapInitialized() throws Exception {
>         def script = runScript("job/params.jenkins")
>         script.execute()
>         super.testNonRegression("params")
>     }
> ```
> 
> ...with the corresponding Jenkinsfile:
> 
> ```groovy
> def execute() {
>     echo params.getClass().toString()
>     echo(params.FOO)
> }
> return this
> ```
> 
> The test fails, even if this patch is applied (condensed stacktrace):
> 
> ```
> class java.lang.Class
> 
> groovy.lang.MissingPropertyException: No such property: FOO for class: params
> 
> 	at groovy.lang.MetaClassImpl.invokeStaticMissingProperty(MetaClassImpl.java:1001)
>         ...
> 	at ___cps.transform___(Native Method)
> 	at com.lesfurets.jenkins.unit.cps.PipelineTestHelperCPS.callMethod(PipelineTestHelperCPS.groovy:32)
>         ...
> 	at com.lesfurets.jenkins.TestRegression.testParamsMapInitialized(TestRegression.groovy:33)
>         ...
> 	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:53)
> ```
> 
> Note that the echo statement thinks that `params` is a `Class` object, not a `Map`.

The issue is that calling `addParam('FOO','x')` would throw an exception unless you first initialize `params` to `[:]` as you do with `env`

Your test fails with/without the patch, as it is testing something different. If you were to try and have your test pass like so:
```groovy
     @Test
     void testParamsMapInitialized() throws Exception {
         def script = runScript("job/params.jenkins")
         addParam('FOO','x')
         script.execute()
         super.testNonRegression("params")
     }
 ```
 without this patch, this test would fail as you will get exception when calling `addParam`

stchar:
I think this fix is outdated as `master` already has a case to handle empty `params`
https://github.com/jenkinsci/JenkinsPipelineUnit/blob/master/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy#L435-L438

stchar:
Will be released in v1.9

PR #322, opened 2020-11-24
Adding unzip mock

cwiggs:
This commit adds a mocked method for 'unzip'.  This will help when someone needs to use the unzip step, making it so they don't have to mock it for the test to succeed.

I used the writeFile mock as a template since it does something similar to unzip.


nre-ableton:
Looks great, thanks for the patch!

nre-ableton (approved):

PR #326, opened 2020-12-17
Fix WhenDeclaration environment condition.

### LINKED ISSUE #325: WhenDeclaration environment condition isn't working properly
https://github.com/jenkinsci/JenkinsPipelineUnit/blob/a469daf904e8c8e9957dea6e2eea9c59150fd9e2/src/main/groovy/com/lesfurets/jenkins/unit/declarative/WhenDeclaration.groovy#L74-L79

The WhenDeclaration implementation of the envCheck doesn't appear to be working properly. Right now it appears to be checking that the environment variable of each entry's key is equal to its value. Instead I think it needs to support the 'name' and 'value' parameters and ensure that the value of the specified environment variable matches the expected value. See https://www.jenkins.io/doc/book/pipeline/syntax/#when for more information.

We can verify it is working properly by updating the current TestDeclarativePipeline.groovy code to assert that the 'Checkout' stage is not run (the environment variable CC is set to 'gcc' and the checkout stage runs when the CC env var is set to 'clang'). See: https://github.com/jenkinsci/JenkinsPipelineUnit/blob/a469daf904e8c8e9957dea6e2eea9c59150fd9e2/src/test/jenkins/jenkinsfiles/Declarative_Jenkinsfile#L23

pkwarren:
Update the WhenDeclaration environment condition to support the 'name'
and 'value' parameters. Fix the logic of evaluating the condition to
ensure that the environment variable matches the expected value. Update
tests to verify the environment condition is now being evaluated
properly.

Fixes #325.

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/declarative/WhenDeclaration.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/declarative/WhenDeclaration.groovy
@@ -14,6 +14,8 @@ class WhenDeclaration extends GenericPipelineDeclaration {
     String branch
     String tag
     Closure<Boolean> expression
+    String envName
+    String envValue
 
     private static Pattern getPatternFromGlob(String glob) {
         // from https://stackoverflow.com/a/3619098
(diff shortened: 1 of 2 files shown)


stchar (approved):
Looks great

stchar:
Merged, thank tou

PR #334, opened 2021-01-15
Fix the buildingTag() method's support

### LINKED ISSUE #330: Unable to validate a declarative conditional using the "buildingTag()" method
Hello there 👋 

Following the PR #220, I was expecting to be allowed to test declarative pipeline with a conditional stage using the `buildingTag()` condition.

However, it looks like there is a bug which results in the condition never triggered (always false), while using the condition `tag '*'` works as expected.

I was able to reproduce this issue with the unit test harness:

- The file [TestDeclarativePipeline.groovy](https://github.com/jenkinsci/JenkinsPipelineUnit/blob/master/src/test/groovy/com/lesfurets/jenkins/unit/declarative/TestDeclarativePipeline.groovy) has 2 tests ([`when_tag` - L176](https://github.com/jenkinsci/JenkinsPipelineUnit/blob/master/src/test/groovy/com/lesfurets/jenkins/unit/declarative/TestDeclarativePipeline.groovy#L176) and [`when_tag_not` - L184](https://github.com/jenkinsci/JenkinsPipelineUnit/blob/master/src/test/groovy/com/lesfurets/jenkins/unit/declarative/TestDeclarativePipeline.groovy#L184)) around the "triggering tag" conditions, but both tests are relying on the `tag` condition.

- If you edit the L11 of the pipeline file [Tag_Jenkinsfile](https://github.com/jenkinsci/JenkinsPipelineUnit/blob/master/src/test/jenkins/jenkinsfiles/Tag_Jenkinsfile), used as fixture for these tests,  to replace `tag 'v*'` by `buildingTag()`, then:
  - `when_tag` is expected to be still successful (whatever the tag, we expect the stage to be executed), but it fails
  - `when_tag_not` is expected to fail (the stage is not skipped anymore) but it succeeds

=> My skills are still not good enough to find why do we have this behavior. I'll try to hack around to understand the code, but any help is welcome...

Thanks a lot for this great tool!

dduportal:
This PR fixes #330.

It introduces the following changes:

- Add 1 test case to validate that the method `buildingTag()` works as expected when testing
- Add 1 test case to cross-verify the "no tag" case with both `tag` and `buildingTag()` instructions
- Fix the bug related to a mis-usage of the containsKey() method

Thanks for this great project folks! Do not hesitate to criticize the content of this PR as I'm learning a bit more Groovy by reading this codebase and trying to fix the bug, so I might do terrible things...

Signed-off-by: dduportal <damien.duportal@gmail.com>

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/declarative/WhenDeclaration.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/declarative/WhenDeclaration.groovy
@@ -10,7 +10,7 @@ class WhenDeclaration extends GenericPipelineDeclaration {
 
     AnyOfDeclaration anyOf
     NotDeclaration not
-    Boolean buildingTag = false
+    Boolean buildingTag
     String branch
     String tag
     Closure<Boolean> expression
@@ -66,7 +66,7 @@ class WhenDeclaration extends GenericPipelineDeclaration {
             branchCheck = antPathMatcher.match(branch, delegate.env.BRANCH_NAME)
         }
         if (buildingTag) {
-            tagCheck = delegate.env.containsKey(TAG_NAME)
+            tagCheck = delegate?.env?.containsKey("TAG_NAME")
         }
         if (tag) {
             tagCheck = delegate.env.TAG_NAME =~ tag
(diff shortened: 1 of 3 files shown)


bitwiseman (approved):

stchar (approved):
Looks good

PR #339, opened 2021-02-16
Remove default values from git plugin example

MarkEWaite:
Default values for submodule configuration generation are ignored in git plugin 4.6.0.  Changes from the default values are untested and expected to be unworkable in earlier plugin versions.

Remove those settings to reduce "noise"


nre-ableton (approved):
Thanks for the fix!

PR #341, opened 2021-02-16
Remove obsolete doGenerateSubmoduleConfigurations

MarkEWaite:
Git plugin versions prior to 4.6.0 included an experiment in submodule configuration that was intentionally excluded from the user interface.  However, the Pipeline uses introspection to identify the DataBoundConstructor and its arguments.  Because that experiment was part of the arguments to the constructor, they were made visible to pipeline users through the Pipeline Syntax snippet generator.

Git plugin 4.6.0 removes those experimental arguments from the snippet generator.  Would be good to also remove them from other source code when we can.  The arguments are still supported, but the values assigned to those arguments are ignored.


nre-ableton (approved):
Thanks!

PR #354, opened 2021-03-24
Add Kubernetes and file operations steps mocks

mahmoud-ashi:
Add mocks for steps from https://github.com/jenkinsci/kubernetes-plugin
Add mocks for steps from https://github.com/jenkinsci/file-operations-plugin
Add mock for `libraryResource`, `lock`, `powershell`
Fix mock for `unstash` step
Fix recently added mock for `timeout` with integer argument

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -136,6 +136,7 @@ abstract class BasePipelineTest {
         helper.registerAllowedMethod("choice", [Map])
         helper.registerAllowedMethod('cifsPublisher', [Map], {true})
         helper.registerAllowedMethod('cleanWs')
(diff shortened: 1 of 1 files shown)


nre-ableton (changes requested):
Please split 963ce5724685ecea433c5da62454c181d70d7e1f up into 5 commits:

- One for the `libraryResource` mock
- One for the `lock` mock
- One for the `powershell` mock
- One for the `timeout` fix
- One for the `unstash` fix (which should probably include a corresponding fix for `stash`, since these steps are related).

That will make this PR much easier to review.

As for the mocks for Kubernetes and file-operations-plugins, please also split those into separate commits and make new PRs for each of them. I am personally disinclined to include mocks for such plugins in this library, since I believe that it should only cover base pipeline steps (otherwise, there is no limit to how much this library will grow, and it will get more difficult to maintain). However, I may be convinced otherwise, but we should have these discussions with other maintainers/contributors in a separate PR.

Thanks!

PR #358, opened 2021-04-01
Avoid using default charset encoding for writing callstack file

reinholdfuereder:
On Windows uses Windows-1252 leading to problems for e.g. German Umlaute


nre-ableton (approved):
Thanks for the fix!

PR #359, opened 2021-04-09
Fix README for "Mocking sh"

reinholdfuereder:
Especially with respect to broken example for executing a closure: must return a map


nre-ableton (approved):
Thanks!

PR #360, opened 2021-04-27
Avoid using default encoding

TurboLuke:
On Windows uses Windows-1252 leading to problems for e.g. German Umlaute

See incomplete #358


nre-ableton (approved):
FYI the commonly-used plural of umlaut is umlauts. Wiktionary does [have an entry for umlaute](https://en.wiktionary.org/wiki/umlaute#English), but it notes that it is a "rare, Germanism". I wasn't sure if "umlaute" was correct at all so I had to look it up. :smile: As the commit message is technically correct, I am not insisting on a change, but just wanted to let you know. :wink:

nre-ableton:
Thanks for the fix!

PR #422, opened 2021-11-11
Add support for mocking library with mapped args

### LINKED ISSUE #418: Mock `library` that takes map args
In https://github.com/opensearch-project/opensearch-build/pull/924 I am loading a library via local (`legacySCM`) SCM.

```
lib = library(identifier: "jenkins@current", retriever: legacySCM(scm))
``` 

I needed to stub `scm`, `legacySCM` and `library`.

```groovy
        binding.setVariable('scm', {})
        helper.registerAllowedMethod("legacySCM", [Closure.class], null)
        
        helper.registerAllowedMethod("library", [Map.class], { Map args ->
            helper.getLibLoader().loadLibrary(args["identifier"])
            return new LibClassLoader(helper, null)
        })
```

Does this belong in the library? We probably need to do more and pass SCM through? What's the right way to do this?

nre-ableton:
This PR adds support for mocking the `library` step when called with mapped arguments.

Fixes #418, FYI @dblock 

- [X] Link to relevant issues in GitHub or Jira
- [X] Link to relevant pull requests, esp. upstream and downstream changes
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue


dblock:
Amazing! Thanks

nre-ableton:
@dblock regarding the `scm` and `legacySCM` parts of your original issue, is this PR enough? Or does it need something else to also cover those?

src/test/jenkins/job/library/libraryJob_dynamic_map.jenkins
```diff
@@ -0,0 +1,10 @@
+def execute() {
+    node() {
+        def lib = library(identifier: 'commons', changelog: false)
+        def utils = lib.net.courtanet.jenkins.Utils.new()
+        sh "${utils.gitTools()}"
+        sh 'sleep 3'
```
dblock, line 6: What’s this sleep about? :)

dblock (approved):

src/test/jenkins/job/library/libraryJob_dynamic_map.jenkins
nre-ableton: I dunno, I just copy/pasted it from another library test. :see_no_evil: However, let's keep things simple; I'll remove the line (see fixup).

dblock:
> @dblock regarding the `scm` and `legacySCM` parts of your original issue, is this PR enough? Or does it need something else to also cover those?

I think it’s definitely enough and we can add later the other stubs. I am not super familiar whether one expects `scm` or `legacySCM` to always be defined. If so we probably want those too.

nre-ableton:
> > @dblock regarding the `scm` and `legacySCM` parts of your original issue, is this PR enough? Or does it need something else to also cover those?
> 
> I think it’s definitely enough and we can add later the other stubs. I am not super familiar whether one expects `scm` or `legacySCM` to always be defined. If so we probably want those too.

`scm` should (and is) defined. I have no idea about `legacySCM` -- I don't find it in the pipeline steps docs, nor in the snippet generator. Let's ignore it out for now. In the meantime, I'll squash and merge this branch, and probably prepare a release sometime next week.

PR #444, opened 2021-11-19
Fix: executing on agent message logs a binding value.

### LINKED ISSUE #433: Executing on agent message logs a binding value
### Version report

1.12

### Reproduction steps

Trying to do a regression test with:

```
agent {
  docker {
    label '...'
    image '...'
  }
}
```

This logs "Executing on agent ..." with a dynamic `binding:groovy.lang.Binding@4822f2cd` value, which changes every run making it difficult to write regression tests with.

```
Build_Jenkinsfile.echo(Executing on agent [docker:[args:TODO, containerPerStageRoot:false, label:Jenkins-Agent-al2-arm64-c6g4xlarge-Docker-Host, image:TODO, reuseNode:false, binding:groovy.lang.Binding@4822f2cd], stages:[:]]])
        ...> but was:<...groovy.lang.Binding@[60a329a7, stages:[:]]])
```

dblock:
Fixes #433 by also removing `binding` parameters. This seems a bit hacky, open to better suggestions.

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/declarative/ObjectUtils.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/declarative/ObjectUtils.groovy
@@ -3,8 +3,9 @@ package com.lesfurets.jenkins.unit.declarative
 class ObjectUtils {
 
     static String printNonNullProperties(Object obj) {
-        def props = obj.properties
+        def props = obj.properties.clone()
         props.remove('class')
+        props.remove('binding')
         obj.properties.entrySet().forEach { e ->
             if (e.value == null) {
                 props.remove(e.key)
--- a/src/test/groovy/com/lesfurets/jenkins/unit/declarative/TestDockerAgentInStep.groovy
+++ b/src/test/groovy/com/lesfurets/jenkins/unit/declarative/TestDockerAgentInStep.groovy
@@ -24,4 +24,12 @@ class TestDockerAgentInStep extends DeclarativePipelineTest {
         runScript("Dockerfile_agent_JenkinsFile")
         assertJobStatusSuccess()
     }
+
+    @Test
+    void test_dockerfile_agent_callstack_does_not_contain_binding() {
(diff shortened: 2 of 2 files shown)


src/main/groovy/com/lesfurets/jenkins/unit/declarative/ObjectUtils.groovy
```diff
@@ -3,8 +3,9 @@ package com.lesfurets.jenkins.unit.declarative
 class ObjectUtils {
 
     static String printNonNullProperties(Object obj) {
-        def props = obj.properties
+        def props = obj.properties.clone()
```
dblock, line 6: I think we need to `.clone()` here, aren't we getting a reference in `obj.properties`?

dblock:
@nre-ableton bumping myself on top of your list - want some help?

nre-ableton (commented):
Sorry for the delay. The code mostly looks fine, but I find the commit message confusing. First, it would be nice if there was some explanation in the commit message body explaining _why_ we shouldn't be logging the binding. It's not very obvious to me why we don't want to have this or what problems it may cause.

Second, the test case that you added isn't very convincing. Wouldn't it be better to assert that the callstack frame for the binding _isn't_ present?

dblock:
I amended the commit message and added a `assertCallStack().doesNotContain('binding:groovy.lang.Binding@')` to the test. I think it's OK to leave the check for the complete expected message in place to avoid any other regressions, but let me know if you want more changes!

nre-ableton (approved):
Thanks, this looks much better!

nre-ableton:
> I think it's OK to leave the check for the complete expected message in place to avoid any other regressions...

I agree.

dblock:
Care to cut a release @nre-ableton, please? Been trying to add some more tests to our complete pipeline jobs and this is the only blocker :(

nre-ableton:
@dblock Sure, I can do that tomorrow

PR #453, opened 2021-11-25
Add support for branch/tag comparators in when block

### LINKED ISSUE #452: Add support for branch/tag comparators in when block
### Feature Request

https://www.jenkins.io/doc/book/pipeline/syntax/#when

As per the Jenkins declarative pipeline syntax, branch and tag conditions in a given when block should be able to support EQUALS, GLOB, and REGEXP comparators. If only a string is passed to these functions, then the GLOB comparator is used implicitly. This seems to be the only functionality supported by the JenkinsPipelineUnit library at the moment. I propose that support be added for passing a map to these functions, so that both pattern and comparator can be provided.

Alex-Weatherhead:
This pull-requests adds support for providing comparators for branch and tag filters in when blocks. As part of this work, functionality from the ChangeRequestDeclaration was separated out into the parent class (WhenDeclaration) to ensure that the approach was consistent with the similar work that had been done there.

Fixes #452.

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/declarative/AllOfDeclaration.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/declarative/AllOfDeclaration.groovy
@@ -1,18 +1,41 @@
 package com.lesfurets.jenkins.unit.declarative
 
-import org.springframework.util.AntPathMatcher
-
 import static groovy.lang.Closure.DELEGATE_FIRST
 
 class AllOfDeclaration extends WhenDeclaration {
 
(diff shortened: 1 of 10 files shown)


BoydL:
Can this be looked into I am experiencing an issue with using pattern matching for my tags

nre-ableton (approved):
I'm quite clueless about the declarative API and tend to stay away from issues relating to that in this repository, but you seem to at least have provided some tests etc. Beyond that, I'm unable to give a thorough review to say if something has been overlooked or missed.

PR #460, opened 2021-12-22
Use the same stringInterceptor in declarative and scripted pipelines

sams-gleb:
Basically just move the string method onto BasePipeline class as params could be used in the scripted pipeline, not only in declarative

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -23,7 +23,22 @@ abstract class BasePipelineTest {
 
     ClassLoader baseClassLoader = this.class.classLoader
 
-    def stringInterceptor = { m -> m.variable }
+    def paramInterceptor = { Map desc ->
+        addParam(desc.name, desc.defaultValue, false)
+    }
+
+    def stringInterceptor = { Map desc->
+        if (desc) {
+            // we are in context of parameters { string(...)}
+            if (desc.name) {
+                addParam(desc.name, desc.defaultValue, false)
+            }
+            // we are in context of withCredentials([string()..]) { }
+            if(desc.variable) {
+                return desc.variable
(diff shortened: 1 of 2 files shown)


nre-ableton (approved):
Nice, thanks for the fix.

reinholdfuereder:
@nre-ableton I think the approach from https://github.com/jenkinsci/JenkinsPipelineUnit/pull/298#issuecomment-811802180 for "stringInterceptor" in combination with "parameters" and "withCredentials" steps (both having nested "string" steps) would be superior:

https://github.com/reinholdfuereder/JenkinsPipelineUnit/tree/parameters
- string parameter: https://github.com/reinholdfuereder/JenkinsPipelineUnit/commit/08589996302001c11f6152c25173315a9f3e925d
- string parameter AND credential: https://github.com/reinholdfuereder/JenkinsPipelineUnit/commit/d77d261ceeb49e41481991ba6fbe214df0b9c51b

nre-ableton:
@reinholdfeurerder I really have no idea what is going on in #298, which is why I didn't review it. In general, I'm not familiar with declarative pipelines, so I tend to stay away from that part of the codebase.

However, that PR doesn't make clear what the problem is nor why this is a good solution.

If you would like to submit a new PR with a similar solution that can better communicate the situation, I'd be happy to review it.

reinholdfuereder:
After starting with a new branch for a new PR I must admit that I re-evaluated the situation and must admit that I have to revoke my claim:
* the advantage I had thought was superior is also a disadvantage, because the "parameters" element than contains duplicates of all the actual parameters (that are in the callstack right above)
   * in the following screenshot on the left my special overriding of "string" interceptor (and non-default "parameters" interceptor); and on the right after reverting them and using default "string" interceptor of Jenkins Pipeline Unit v1.14 (plus an additional default "parameters" interceptor => https://github.com/jenkinsci/JenkinsPipelineUnit/pull/495 with that and some new unit tests)

image

PR #425, opened 2021-11-11
Replace AppVeyor and Travis-CI with GitHub Actions.

dblock:
Replaces Travis and AppVeyor CI by GitHub Actions where we can run CI for all platforms. AppVeyor and Travis can/should be turned off before merge in repo settings (I think there are permissions for those that should be removed).

Passing build: https://github.com/dblock/JenkinsPipelineUnit/actions/runs/1449798899

Closes https://github.com/jenkinsci/JenkinsPipelineUnit/issues/424

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #482, opened 2022-02-09
Feature: Run tests in Github Actions.

UlrichEckhardt:
GitHub Actions are GitHub's CI/CD feature. This pull request just adds a very simple action which runs the existing tests. The run is triggered by a commit to any branch and then executes the tests inside a Docker container.

I haven't added that yet, but you can use this to generate badges. You can then integrate them into e.g. the readme.md, like for example the Go tests in https://github.com/UlrichEckhardt/api-broker-prototype#readme.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -1,0 +1,13 @@
+name: Gradle Tests
+on: [push]
+jobs:
+  gradle_test:
+    name: Run tests with Gradle
+    runs-on: ubuntu-latest
+    container: gradle:6.6-jre14-hotspot
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Run Gradle tests
+      run: gradle --console plain


nre-ableton:
@UlrichEckhardt Yes, solving #425 has been on my to-do list for awhile. :sweat_smile: Anyways, I asked around the Jenkins infrastructure IRC channels as to what the preferred approach would be, and the general answer is "anything is better than what we have now". :smile:

I would prefer to use Jenkins' official open-source service, given that this project is for Jenkins anyways, but I couldn't figure out why it wasn't running our builds. My next preferred solution would be GitHub actions, so I would be happy to at least upgrade from AppVeyor to this.

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,13 @@
+name: Gradle Tests
+on: [push]
```
nre-ableton, line 2: Can we also not run CI for PRs? Ideally, I'd like to _replace_ AppVeyor with GH actions entirely.

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,13 @@
...
+  gradle_test:
+    name: Run tests with Gradle
+    runs-on: ubuntu-latest
+    container: gradle:6.6-jre14-hotspot
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Run Gradle tests
+      run: gradle --console plain
```
nre-ableton, line 13: So on AppVeyor, we currently do:

```
gradlew.bat --no-daemon assemble
gradlew.bat --no-daemon check
```

I'm not sure where you found `gradle --console plain`, and I'm no gradle wizard myself, so I'm not 100% sure what that does... but I did try it locally and it seemed to run the tests. :shrug: Where did this come from?

nre-ableton:
Hmm, sorry for the noise, but having another look at #425, maybe we should go with that one instead since it already includes a lot of the feedback I gave here. Sorry!

.github/workflows/gradle.yml
UlrichEckhardt: `--console plain` is to remove any animations that are aimed at the terminal but not at a logfile.

Apart from that, I disliked the use of gradlew/gradlew.bat in a different project (putting generated code into version control), which is why I searched for a way to run this inside a Docker container. Actually, you don't need those two wrapper scripts then. Also, locally, I keep a running Docker container just for the daemon to speed up subsequent tests.

.github/workflows/gradle.yml
UlrichEckhardt: I think that once you have it merged, any fork automatically gets these actions, too. So they will automatically run the tests on every commit they make, even before making a PR. I think that's what you're after, right?

nre-ableton: Kind of. On the one hand, I was confused why GitHub actions weren't present for this PR. I guess we have to merge to `master` first. However, looking at #425, I guess it should [include](https://github.com/jenkinsci/JenkinsPipelineUnit/pull/425/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR3) `pull_requests` as well.

nre-ableton:
@UlrichEckhardt Do you mind if I close this PR?

dduportal:
> 

Hello 👋 Jenkins Infra team meber here: if you want the project to have it's CI on the public ci.jenkins.io, it's totally feasible. We've improved its stabilit since last year, it might be intetesting to reuse the same build farm system as all Jenkins plugins are doing?

However GitHub actions also makes sense of course, if it's easier to use or more reliable for your usage.

Do not hesitate to open an issue in github.com/jenkins-infra/helpdesk to describe your needs if interested :)

nre-ableton:
@dduportal Yes, I would prefer to use Jenkins since as you mentioned, that's what the plugins use and also, this framework is already pretty Jenkins-centric. :smile: I'll make an issue there.

UlrichEckhardt:
Yes, I think this one can be closed.

PR #483, opened 2022-02-09
Fix sh default output

UlrichEckhardt:
The mocked sh behaviour will return a string for unconfigured scripts. If I have for example
```
def x = sh(script: 'something')
def y = sh(script: 'something', returnStdout: true)
```
`x` will be `null`, only `y` will receive a value in Jenkins, but with the test framework both will receive a value. See also https://www.jenkins.io/doc/pipeline/steps/workflow-durable-task-step/#sh-shell-script. This misbehaviour is manifested in a test even. The provided PR fixes the behaviour of the mock and replaces the faulty test with two operational tests.


nre-ableton (changes requested):
I've always disliked the default output for both `bat` and `sh` in this library (as you can tell from the comment on lines 714-716, which I wrote).

I think we should go ahead and make the breaking change. Since you've started the effort, let's do it in this PR. Please remove all of this backwards compatibility junk and just return `null` here. In other words, it could be simplified as far as this:

```groovy
        MockScriptOutput output = mockScriptOutputs[script]
        if (!output) {
            return returnStatus ? 0 : null
        }
```

Let's also make a similar change for `bat` in this PR. After rebasing, that should make your other PR #484 a bit simpler.

I will take care of the changelog and fallout from anyone complaining about the API breakage (assuming there is anyone out there who still depends on this default behavior -- I somewhat doubt anybody is, though).

The changes to the test case look fine.

Thanks for your work!

nre-ableton (approved):
Perfect, thanks! 👍

PR #510, opened 2022-03-31
Generate a default closure in InterceptingGCL

nestoracunablanco:
In case a method is matched with null value closure, the default generated closure is {}.

This does not work when the method has more parameters, so the approach of generating a default closure using dynamic compilation was taken.

- [X] Link to relevant issues in GitHub or Jira
- [X] Link to relevant pull requests, esp. upstream and downstream changes
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue

Solves comments in issue #461

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
@@ -18,9 +18,18 @@ class InterceptingGCL extends GroovyClassLoader {
             Map.Entry<MethodSignature, Closure> matchingMethod = helper.allowedMethodCallbacks.find { k, v -> k == signature }
             if (matchingMethod) {
                 // a matching method was registered, replace script method execution call with the registered closure (mock)
-                metaClazz."$scriptMethod.name" = matchingMethod.value ?: {}
+                metaClazz."$scriptMethod.name" = matchingMethod.value ?: defaultClosure(matchingMethod.key.args)
             }
         }
+    }
+
+    static Closure defaultClosure(Class[] args) {
+        String argumentsString = args.inject("") { acc, value ->
+            acc + value.name + " " + "a"*(acc.count(',') + 1) + ","}
+        String argumentsStringWithoutComma = argumentsString.size() > 0 ?
+                argumentsString.substring(0, argumentsString.length()-1) : argumentsString
+        String closureString = "{$argumentsStringWithoutComma -> }"
+        return (Closure)new GroovyShell().evaluate(closureString)
     }
 
     PipelineTestHelper helper
--- a/src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
@@ -24,8 +24,10 @@ class InterceptingGCL extends GroovyClassLoader {
     }
 
     static Closure defaultClosure(Class[] args) {
+        int maxLength = 254
+        if (args.length > maxLength) { throw new IllegalArgumentException("Only $maxLength arguments allowed")}
         String argumentsString = args.inject("") { acc, value ->
-            acc + value.name + " " + "a"*(acc.count(',') + 1) + ","}
+            return "${acc}${value.name} ${'a'*(acc.count(',') + 1)},"}
         String argumentsStringWithoutComma = argumentsString.size() > 0 ?
                 argumentsString.substring(0, argumentsString.length()-1) : argumentsString
         String closureString = "{$argumentsStringWithoutComma -> }"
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
```diff
@@ -18,11 +18,20 @@ class InterceptingGCL extends GroovyClassLoader {
...
                 // a matching method was registered, replace script method execution call with the registered closure (mock)
-                metaClazz."$scriptMethod.name" = matchingMethod.value ?: {}
+                metaClazz."$scriptMethod.name" = matchingMethod.value ?: defaultClosure(matchingMethod.key.args)
             }
         }
     }
 
+    static Closure defaultClosure(Class[] args) {
+        String argumentsString = args.inject("") { acc, value ->
+            acc + value.name + " " + "a"*(acc.count(',') + 1) + ","}
```
nre-ableton, line 28: IMHO this is more readable:

```suggestion
        String argumentsString = args.inject("") { acc, value ->
            return "${acc}${value.name} a${*(acc.count(',') + 1)},"
        }
```

src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
```diff
@@ -18,11 +18,20 @@ class InterceptingGCL extends GroovyClassLoader {
...
+        String argumentsStringWithoutComma = argumentsString.size() > 0 ?
+                argumentsString.substring(0, argumentsString.length()-1) : argumentsString
+        String closureString = "{$argumentsStringWithoutComma -> }"
+        return (Closure)new GroovyShell().evaluate(closureString)
```
nre-ableton, line 32: I don't think a cast is the way to go here. Why not return a real closure?

```suggestion
        return { new GroovyShell().evaluate(closureString) }
```

nre-ableton (changes requested):
There is a lot of logic going on in `defaultClosure` that isn't tested. It would be nice to have a unit test to make sure that all of the string slicing and appending works as expected.

Also, there is a typo in the commit message (it should be "_Generate_ a default ..."). I fixed this in the PR title for you.

src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
nestoracunablanco: This change did not work out of the box, instead I did something similar. I hope I got the idea behind your suggestion.

src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
nestoracunablanco: I agree with you, that would be nicer... but this did not work. Testing the suggested way, it created a closure with a closure inside, which is not intended.

nestoracunablanco:
@nre-ableton more unit tests are now in place.

src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
```diff
@@ -18,11 +18,22 @@ class InterceptingGCL extends GroovyClassLoader {
...
+                metaClazz."$scriptMethod.name" = matchingMethod.value ?: defaultClosure(matchingMethod.key.args)
             }
         }
     }
 
+    static Closure defaultClosure(Class[] args) {
+        int maxLength = 254
+        if (args.length > maxLength) { throw new IllegalArgumentException("Only $maxLength arguments allowed")}
+        String argumentsString = args.inject("") { acc, value ->
+            return "${acc}${value.name} ${'a'*(acc.count(',') + 1)},"}
```
nre-ableton, line 30:
```suggestion
        if (args.length > maxLength) {
            throw new IllegalArgumentException("Only $maxLength arguments allowed")
        }
        String argumentsString = args.inject("") { acc, value ->
            return "${acc}${value.name} ${'a'*(acc.count(',') + 1)},"
        }
```

...just a minor suggestion to improve readability.

nre-ableton (approved):
I guess this change looks fine to me, but I am not sure I can speak to the broader impact that this change may have. It would be nice if @stchar could also take a look.

nestoracunablanco:
Hi @stchar, can you please also take a look at this pull request? Thanks!

stchar (approved):
Awesome!

stchar:
@nestoracunablanco, @nre-ableton, reviewed. Looks good. 
I would also add tests for scripted pipeline, but I assume there is no any difference as `InterceptingGCL` address to metaclass of `groovy.lang.Script` which is the same in case of Declarative and Scripted pipeline test runs.

nestoracunablanco:
Thanks for the nice feedback!

Issue #478, opened 2022-02-04
Params not accessible in SharedLibrary classes


MichaelJKar:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
  See minimal viable example below. Plugins should not play a role here.
  ```

</details>

### What Operating System are you using (both controller, and any agents involved in the problem)?

Windows

### Reproduction steps

**Jenkinsfile:**
```
#!groovy
@Library('lib')
import LibClass

pipeline {
    stages {
        stage('Test stage'){
            steps {
                out = new LibClass().getX()
                echo "Printing ${out}"
            }
        }
    }
}
```

**Test Class:**
```
import com.lesfurets.jenkins.unit.declarative.DeclarativePipelineTest
import org.junit.jupiter.api.BeforeEach
import org.junit.jupiter.api.Test

import static com.lesfurets.jenkins.unit.global.lib.LibraryConfiguration.library
import static com.lesfurets.jenkins.unit.global.lib.ProjectSource.projectSource

class MainPipelineIntegrationTest extends DeclarativePipelineTest {

	@Override
	@BeforeEach
	void setUp() throws Exception {
		super.setUp()
		def library = library().name("lib")
				.retriever(projectSource("./mocklib"))
				.defaultVersion("master")
				.targetPath("target/mocklib")
				.allowOverride(true)
				.implicit(false)
				.build()
		helper.registerSharedLibrary(library)
	}

	@Test
	void test() {
		binding.setVariable('env', ['X':'testXValue'])
		binding.setVariable('params', ['X':'testXValue'])

		runScript("../Jenkinsfile")

		assertJobStatusSuccess()
		printCallStack()
	}
}
```

**The shared library `lib` only contains one file `LibClass.groovy` with the following content:**
```
class LibClass {
    def getX(){
        return params.X
    }
}
```

### Expected Results

```
Loading shared library lib with version master
Printing testXValue
   Jenkinsfile.run()
      Jenkinsfile.pipeline(groovy.lang.Closure)
         Jenkinsfile.stage(Test stage, groovy.lang.Closure)
            LibClass.getX()
            Jenkinsfile.echo(Printing testXValue)
```

### Actual Results

Loading shared library lib with version master

groovy.lang.MissingPropertyException: No such property: params for class: LibClass

### Anything else?

When replacing the content of `getX()` with `return env.X`, the test runs successfully and the output matches the expected output given above.

[...]

PR #519, opened 2022-04-16
Make bindings accessible in SharedLibrary classes.

nestoracunablanco:
For the library loading a Pogo is instantiated, which has a different implementation while getting an object.

This POGO does not read the bindings. In the PogoMetaClassGetPropertySite the implementation can be found. This throws a runtime exception, which suits well for the common purposes, but not when we try to emulate the Jenkins shared library behaviour.

So the workaround is to add the bindings as properties dynamically inside InterceptingGCL.

This change tries to solve the issue #478 

- [ X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ X] Ensure that the pull request title represents the desired changelog entry
- [ X] Please describe what you did
- [ X] Link to relevant issues in GitHub or Jira
- [ X] Link to relevant pull requests, esp. upstream and downstream changes
- [ X] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
@@ -12,6 +12,9 @@ class InterceptingGCL extends GroovyClassLoader {
         metaClazz.static.invokeMethod = helper.getMethodInterceptor()
         metaClazz.methodMissing = helper.getMethodMissingInterceptor()
         metaClazz.getEnv = {return binding.env}
+        binding.variables.forEach { String property, Object value ->
+            metaClazz."$property" = metaClazz."$property" ?: value
+        }
         // find and replace script method closure with any matching allowed method closure
         metaClazz.methods.forEach { scriptMethod ->
             def signature = method(scriptMethod.name, scriptMethod.nativeParameterTypes)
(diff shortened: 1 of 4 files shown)


src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
```diff
@@ -12,6 +12,9 @@ class InterceptingGCL extends GroovyClassLoader {
         metaClazz.static.invokeMethod = helper.getMethodInterceptor()
         metaClazz.methodMissing = helper.getMethodMissingInterceptor()
         metaClazz.getEnv = {return binding.env}
+        binding.variables.forEach { String property, Object value ->
```
stchar, line 15: @nestoracunablanco, firs of all thank you for the Pull request.

Did you try to use propertyMissing interceptor? 
I'm corious what if you create an instance of library class and then change the binding. 
Could you please add a test to this scenario?

stchar (changes requested):
Please add on more test scenario.

src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
nestoracunablanco: Hi @stchar , thanks for the nice feedback :)
To your question, yes, I tried to use `propertyMissing` interceptor, but since the instantiation relies on `PogoMetaClassGetPropertySite` instead of a `PropertyMissingException` a `GroovyRuntimeException` is thrown. I do not know if this is an intentional design decision or a bug, but this is the pragmatic workaround I found.

nestoracunablanco:
Hi @stchar, could you please take a look at my changes? Is that what you requested?
Thanks in advance!

src/main/groovy/com/lesfurets/jenkins/unit/InterceptingGCL.groovy
stchar: Ok, I see. I tried to implement  `propertyMissing` and `propertyInterceptor` few month ago but wasn't succeed. This groovy interceptor magic is hard)) Glad you  made it worked.

stchar (approved):
Looks good. Thank you!

Issue #387, opened 2021-07-07
Bound variables are not available in `environment` closure


pavelsmolensky:
Given following (simplified) library files in vars directory:
testLib.groovy:

```
def call(body) {
    pipeline {
        environment {
            HI = helper.sayHi() //here helper is null
        }
        stages {
            stage("One") {
                helper.sayHi()  //however here helper is defined
                echo "${env.HI}"
            }
        }
    }
}
```

helper.groovy:

```
def sayHi(){
    return 'hi'
}
```

And test class:

```
class DemoTest extends DeclarativePipelineTest {
    @Override
    @BeforeEach
    public void setUp() throws Exception {
        super.setUp();
    }    

    @Test
    public void demo(){
        def lib = loadScript("vars/testLib.groovy")
        def helper = loadScript("vars/helper.groovy")
        binding.setVariable("helper", helper)
        lib.call({})
        printCallStack()
    }
}    
```

There is an exception when line HI = helper.sayHi() is executed in `environment` closure:

```
java.lang.NullPointerException: Cannot invoke method sayHi() on null object
...
	at testLib$_call_closure1$_closure2.doCall(testLib.groovy:4)
...

```
However calling helper.sayHi() at stage level works fine

PR #529, opened 2022-05-29
Fix for shared lib and params variables being null in environment section

tomek-d:

Hello, this is a possible fix for #387.

### Analysis of the issue 
According to Groovy's metaprogramming rules (https://docs.groovy-lang.org/latest/html/documentation/core-metaprogramming.html) Groovy calls `getProperty` method on `Closure` which in turn tries to get this property from the delegate (in our case this is a `Map`) and when the property is not found it tries to find the property on the owner object. 
Here's code snippet from Groovy's source code:
```groovy
    private Object getPropertyTryThese(String property, Object firstTry, Object secondTry) {
        try {
            // let's try getting the property on the first object
            return InvokerHelper.getProperty(firstTry, property);

        } catch (MissingPropertyException | MissingFieldException e1) {
            if (secondTry != null && firstTry != this && firstTry != secondTry) {
                try {
                    // let's try getting the property on the second object
                    return InvokerHelper.getProperty(secondTry, property);
                } catch (GroovyRuntimeException e2) {
                    // ignore, we'll throw e1
                }
            }
            throw e1;

        }
    }
```
https://github.com/apache/groovy/blob/02d1a3b4f76f1311eb2286188b224cfbbf0324db/src/main/java/groovy/lang/Closure.java#L322

In our case delegate is a `Map` and property retrieval for a `Map` is implemented as a call to `get` method on that `Map` as shown on following code snippet from Groovy implementation:
```groovy
if (isMap && !isStatic) {
	return ((Map<?,?>) object).get(name);
}	
```
https://github.com/apache/groovy/blob/master/src/main/java/groovy/lang/MetaClassImpl.java#L2001	

The problem here is that `Map` returns `null` when the property is not found however Closure expects `MissingPropertyException` or `MissingFieldException` exceptions to try to get the property from the owner. 

### Proposed fix
Create a wrapper that first checks if the property is defined in a map and if not then it tries to get property from the owner which in this case is an instance of GenericPipelineDeclaration.

PS. I'm not a Groovy developer so possibly a better solution could be found.

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/declarative/GenericPipelineDeclaration.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/declarative/GenericPipelineDeclaration.groovy
@@ -102,9 +102,37 @@ abstract class GenericPipelineDeclaration {
             env.env = env
             env.currentBuild = delegate.binding.currentBuild
 
-            def cl = this.environment.rehydrate(env, delegate, this)
+            def cl = this.environment.rehydrate(wrapEnv(env as Map<String, Object>), delegate, this)
             cl.resolveStrategy = DELEGATE_FIRST
             cl.call()
+        }
+    }
+
+    private def wrapEnv(Map<String, Object> env) {
+        return new GroovyObject() {
+            Object invokeMethod(String name, Object args) {
+                GenericPipelineDeclaration.this.invokeMethod(name, args)
+            }
+
+            Object getProperty(String propertyName) {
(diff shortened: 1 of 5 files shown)


nre-ableton:
Sorry, for the delay, I haven't had time to work on this project recently, but my team now has a hack sprint so I'll be examining this issue soon.

tomek-d:
@nre-ableton  - thanks for merging this PR ;)

PR #517, opened 2022-04-15
Set currentBuild.currentResult in updateBuildStatus

nre-ableton:
Set `currentBuild.currentResult` when updating build status in `BasePipelineTest.updateBuildStatus`. Fixes https://github.com/jenkinsci/JenkinsPipelineUnit/issues/516.

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Link to relevant issues in GitHub or Jira
- [X] Link to relevant pull requests, esp. upstream and downstream changes
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #605, opened 2023-07-10
BasePipelineTest: Add result to default mock of `build`

reitzig:
Running tests of pipeline scripts that trigger other builds, the build result always ends up being `FAILURE`.

This PR adds `SUCCESS` as default result of triggered builds, which seems reasonable and in the spirit of other mocks such as for `sh`.

I am not aware of any issue tickets that relate to this.
Please advise if any tests are needed for this change.

### Testing done

`BasePipelineTest` has not detailed unit tests.

We've been having this change in use for years via a fork of the library.

### Submitter checklist

- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -137,6 +137,7 @@ abstract class BasePipelineTest {
                 getDescription:{"Dummy build description"},
                 getFullProjectName:{"some_dir/some_job"},
                 getProjectName:{"some_job"},
+                getResult:{"SUCCESS"},
             ]
         })
         helper.registerAllowedMethod('booleanParam', [Map], paramInterceptor)


nre-ableton (approved):
Hi @reitzig! Thanks for your PR. 👍 The changes look fine. If you want to add tests, that would of course be welcome, but this framework doesn't have tests for most trivial mocks (ie, those without any logic).

reitzig:
@nre-ableton thanks for the quick response!

I just did what I should have done before and checked the state of our fork. Turns out it was stuck at 1.13; after bumping to 1.17 and comparing results with between upstream and our fork, it turns out the propsed change is not necessary anymore, at least not to make our specific test suite green. (FWIW, I confirmed that the default mock _is_ being used in the test suite.) Apparently, some behaviour changed so the "silent" default is `SUCCESS`? 🤔

Do you think the change has merit regardless? If not, we can just close the PR.

Sorry for the noise! 🙏

nre-ableton:
I guess that https://github.com/jenkinsci/JenkinsPipelineUnit/pull/517 fixed the issue that you were experiencing.

If the current state is working, I would say that we should rather close this PR.

PR #612, opened 2023-07-28
Keep track of wrapper updates

NotMyFault:
I was looking forward to updating the wrapper to support Java 20, but given it's a sensitive component, I chose to add an action doing that automatically, and validating its integrity.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,7 @@ updates:
     directory: "/"
     schedule:
       interval: daily
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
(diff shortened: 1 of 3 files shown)


nre-ableton:
FYI, I've seen this PR, but am not 100% sure about it yet. Specifically, I wonder if we might be able to use [renovate](https://docs.renovatebot.com/modules/datasource/gradle-version/) for something like this.

NotMyFault:
> FYI, I've seen this PR, but am not 100% sure about it yet. Specifically, I wonder if we might be able to use [renovate](https://docs.renovatebot.com/modules/datasource/gradle-version/) for something like this.

No worries, the decision would be up to you. If you want to use renovate, I'd go ahead and close the PR.
We're using renovate for that already in other repositories, FYI.

PR #620, opened 2023-08-16
Update README.md to discuss setUp

ryeleo:
Make it clear that `super.setUp()` must be called before using test helpers.

Some examples throughout the README omit the call to setUp, so it is nice to discuss it up top, imo.

### Testing done

No testing -- only change to the docs.

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
  > I'm unsure if I'm opening from the proper branch. Should it literally be named "topic/feature/bugfix"?

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,10 +171,14 @@ class TestExampleJob extends BasePipelineTest {
 }
 ```
 
-The test helper already provides basic variables such as a very simple `currentBuild`
-definition. You can redefine them as you wish.
+After calling `super.setUp()`, the test `helper` instance is available, as well as many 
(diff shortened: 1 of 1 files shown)


stefan6419846:
Wouldn't it be better to add the *super* calls to the corresponding examples as well?

nre-ableton:
> Wouldn't it be better to add the *super* calls to the corresponding examples as well?

Yes, I agree that this would be a good idea. Most of the example code does this already but there are a few examples that don't. This PR is pretty stale so I'll merge it as-is and then make a follow-up PR to fix the examples.

PR #662, opened 2024-07-06
docs(README): add environment variable example

### LINKED ISSUE #617: Defining environment variables is not clear
### Describe your use-case which is not covered by existing documentation.

In the documentation inside the README we have: https://github.com/jenkinsci/JenkinsPipelineUnit/blob/c8fa3267c789b9971e5768042c7f70926af4b467/README.md?plain=1#L151-L172

How does this cover setting environment variables? For me, it seems like this will only set a parameter/*params* value, but no *env* one. Playing around with it seems to be indicate that this might be achieved by using `binding.getVariable('env').newValue = 'test'`, but it might make sense to actually document this as well.

### Reference any relevant documentation, other materials or issues/pull requests that can be used for inspiration.

_No response_

axieum:
Add example usage of `addEnvVar` to the README.md examples.

Closes #617 

### Testing done

n/a

### Submitter checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,8 @@ class TestExampleJob extends BasePipelineTest {
         super.setUp()
         // Assigns false to a job parameter ENABLE_TEST_STAGE
         addParam('ENABLE_TEST_STAGE', 'false')
+        // Assigns 1.0.0-rc.1 to the environment variable TAG_NAME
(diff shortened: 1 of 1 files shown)


nre-ableton (commented):
Thanks for your contribution! :+1:

nre-ableton (approved):
Thanks for your contribution! :+1:

Issue #227, opened 2020-05-27
DeclarativePipelineTest - 'when' is missing the option for 'anyOf' and 'allOf'


pcoulso1:
The class WhenDeclaration in com.lesfurets.jenkins.unit.declarative seems to be missing the option for 'anyOf' and 'allOf'

e.g. when used in the stage section of the jenkinsfile
```
stage('build') {
  when {
    allOf {
      environment name: 'cleanBuild', value: 'true'
      expression { BRANCH_NAME ==~ /(develop|hotfix\/.*)/ }
    }
  }
  steps {
    …
  }
}
```	  
And results in the following exception when the tests are run
```
groovy.lang.MissingMethodException: No signature of method: com.lesfurets.jenkins.unit.declarative.WhenDeclaration.allOf() is applicable for argument types: (Jenkinsfile$_run_closure1$_closure6$_closure11$_closure21$_closure24) values: [Jenkinsfile$_run_closure1$_closure6$_closure11$_closure21$_closure24@4149c063]
Possible solutions: any(), sleep(long), any(groovy.lang.Closure), split(groovy.lang.Closure)
            at Jenkinsfile$_run_closure1$_closure6.closure11$_closure21(Jenkinsfile:153)
            at Jenkinsfile$_run_closure1$_closure6.closure11$_closure21(Jenkinsfile)
            at com.lesfurets.jenkins.unit.declarative.DeclarativePipeline.createComponent(DeclarativePipeline.groovy:18)
            at com.lesfurets.jenkins.unit.declarative.StageDeclaration.when(StageDeclaration.groovy:32)
            at Jenkinsfile$_run_closure1.closure6$_closure11(Jenkinsfile:152)
```
See https://www.jenkins.io/doc/book/pipeline/syntax/ for details

lucasvc:
Is there any way to workaround this in my own test code?
I have tried,

```
        testee = loadScript("vars/mypipeline.groovy")
        helper.registerAllowedMethod("anyOf", [Closure.class], { it -> it() })
        testee.call()
```
But still same exception.

Sorry if this is somewhere explained, but I could not find it and I'm totally groovy-newbie.

epsilo:
Hi @IuryAlves 

The implementation of anyOf you made only work for a single validation inside anyOf.
`anyOf { branch 'master' }`
In this case works to check if the branch is master or not.

If we have more than one validation e will only validate the last one
[...]

PR #680, opened 2024-09-13
Added support for beforeAgent property in when statements

MKITConsulting:
This pull request addresses the missing support for `beforeAgent` inside `when` statements, see also issue https://github.com/jenkinsci/JenkinsPipelineUnit/issues/227.

### Testing done

I've added new test cases and a Jenkinsfile with a nested `beforeAgent` property inside a `when` statement to cover the use case that without this PR would lead to the error message mentioned in https://github.com/jenkinsci/JenkinsPipelineUnit/issues/227. 

### Submitter checklist


nre-ableton (approved):
Thanks!

PR #748, opened 2026-02-04
Add Script & Library Caching

The-Jonsey:
Added a script and library cache to the PipelineTestHelper, this helped reduce the time a suite of tests took on a complex pipeline library from several minutes to ~30 seconds, as each test previous took ~7 seconds to load the script and the library. Any changes to mocks or script/libraries during test execution would be ignored by the caching, so the cache is disabled by default with a flag to enable it.

### Submitter checklist
- [ ] Link to relevant issues in GitHub or Jira
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/BasePipelineTest.groovy
@@ -1,5 +1,7 @@
 package com.lesfurets.jenkins.unit
 
+import com.lesfurets.jenkins.unit.global.lib.LibraryLoader
+
 import static java.util.stream.Collectors.joining
 import static org.assertj.core.api.Assertions.assertThat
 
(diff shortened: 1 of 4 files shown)


nre-ableton (changes requested):
First of all, thanks for your PR and sorry for the delay in reviewing. It was indeed in my GitHub notifications, but things have been quite busy at work recently (same old excuses 🙃).

Anyways, the concept of this PR seems fine, but I am not wild about the API. In general, variables/methods/classes/etc with `And` in the name should be an indication that they are trying to do too many things and should be refactored accordingly.

I would recommend the following:

- Remove `shouldCacheScriptsAndLibraries` and replace with two distinct properties: `cacheScripts` and `cacheLibraries`
- Eliminate the `withShouldCacheScriptsAndLibraries` method and simply allow users to set the above properties as usual
- Use the respective fields when clearing cached classes/libraries

I have other minor suggestions, but I don't think they are worth discussing until we figure out the "big picture" here.

PR #771, opened 2026-06-17
Enable parallel testing

juanpablo-santos:
This PR adds support for parallel testing using JUnit. As you can see Claude (Opus 4.8) was used to perform the changes, but there was a human driver behind ;-)

### Testing done

JenkinsPipelineUnit itself runs its tests in parallel (16 seconds on my machine). At work we're using these changes to parallelize the execution of almost 1700 tests on a jenkins commons library and of 500 unit tests associated to our pipelines. We've cut the testing time to a 1/4 of what it used to be.

As the change is big enough, the version is bumped to 2.0.0, although it shouldn't be a breaking change (you've to enable parallelism on your test suite to benefit from it).

### Submitter checklist
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/README.md
+++ b/README.md
@@ -588,6 +588,12 @@ Notes and caveats:
   static maps populated by `@BeforeEach`, and `metaClass` mutation on shared/global objects
   in your test code are not made safe by this framework. Keep such state per-test, or guard
   it.
+- **`@Grab` in pipeline or shared-library code is not safe under parallelism.** Groovy
+  resolves `@Grab` through Grape/Ivy *at compile time*, and Grape uses a process-global,
+  non-thread-safe Ivy instance. Because each helper compiles library classes with its own
+  class loader, two concurrent tests compiling the same `@Grab`-annotated class can corrupt
+  each other's resolution and fail with `unable to resolve class ...`. Pre-resolve such
+  dependencies on the test classpath instead of grabbing them from pipeline code.
 - JenkinsPipelineUnit's own test suite runs concurrently to guard against regressions.
 
 ## Declarative Pipelines
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ and even track regressions.
 
 1. [Usage](#usage)
 1. [Configuration](#configuration)
+1. [Parallel Test Execution](#parallel-test-execution)
 1. [Declarative Pipeline](#declarative-pipelines)
 1. [Testing Shared Libraries](#testing-pipelines-that-use-shared-libraries)
 1. [Writing Testable Libraries](#writing-testable-libraries)
@@ -558,6 +559,36 @@ This will work fine for such a project structure:
          └── groovy
              └── TestExampleJob.groovy
 ```
+
+## Parallel Test Execution
+
+JenkinsPipelineUnit is thread-safe and supports running tests in parallel within a single
+JVM. Each test must use its own `PipelineTestHelper` instance (the default when you subclass
+`BasePipelineTest`/`DeclarativePipelineTest` without sharing a helper): script classes,
+shared-library classes, mocks and the call stack are all isolated per helper, and the
+framework no longer mutates any JVM-global state across helpers.
+
+To enable parallel execution with JUnit 5/6, add a `junit-platform.properties` file to your
+test resources:
+
+```properties
+junit.jupiter.execution.parallel.enabled = true
+junit.jupiter.execution.parallel.mode.default = concurrent
+junit.jupiter.execution.parallel.mode.classes.default = concurrent
+junit.jupiter.execution.parallel.config.strategy = dynamic
+```
+
+Notes and caveats:
+
+- **One helper per test.** If you deliberately *share* a single `PipelineTestHelper`
+  across tests (e.g. a `static` cached helper for speed), that shared helper is still
+  mutable state and must not be used by concurrent tests. Give each concurrently-running
+  test its own helper.
+- **Your own test code must be thread-safe too.** Static mocks (e.g. shared Mockito mocks),
+  static maps populated by `@BeforeEach`, and `metaClass` mutation on shared/global objects
+  in your test code are not made safe by this framework. Keep such state per-test, or guard
+  it.
+- JenkinsPipelineUnit's own test suite runs concurrently to guard against regressions.
 
 ## Declarative Pipelines
 
--- a/src/main/groovy/com/lesfurets/jenkins/unit/DockerMock.groovy
+++ b/src/main/groovy/com/lesfurets/jenkins/unit/DockerMock.groovy
@@ -3,6 +3,33 @@ package com.lesfurets.jenkins.unit
 
 @SuppressWarnings(['EmptyMethod', 'MethodReturnTypeRequired', 'UnusedMethodParameter'])
 class DockerMock implements Serializable {
+
+    // Helper and binding used to apply per-instance interception to the Image/Container
+    // objects this mock creates, so their calls are recorded on the owning helper's call stack.
+    // Transient: these are test-infra references and must not be carried by CPS serialization.
+    private transient PipelineTestHelper helper
+    private transient Binding binding
+
+    DockerMock() {
+    }
+
+    DockerMock(PipelineTestHelper helper, Binding binding) {
+        this.helper = helper
+        this.binding = binding
+    }
+
+    /**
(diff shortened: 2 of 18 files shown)


juanpablo-santos:
~Not sure on how to see the failures on the associated workflows, build runs ok on my machine and the modified library has been using for almost a month without problems, so would appreciate a pointer on how to do it~

Edit: seen https://github.com/jenkinsci/JenkinsPipelineUnit/pull/771/checks with the failing tests, but would appreciate how to see the jenkins logs, can't reproduce the failures on my local machine

thx in advance

lemeurherve:
> Edit: seen https://github.com/jenkinsci/JenkinsPipelineUnit/pull/771/checks with the failing tests, but would appreciate how to see the jenkins logs, can't reproduce the failures on my local machine

To be able to see the failing logs, you have to create an account on https://accounts.jenkins.io then use those credentials to login on ci.jenkins.io to consult https://ci.jenkins.io/job/jenkinsci-libraries/job/JenkinsPipelineUnit/job/PR-771/ builds.

juanpablo-santos:
@lemeurherve thanks for the pointer :-) pushed a new commit which seems to fix the test failures, hopefully the PR is reviewable now

nre-ableton (changes requested):
First of all, thanks for your contribution!

Initially, these changes seem OK but I will take a closer look at them next week. The only change I would request right now is not to bump the JenkinsPipelineUnit version (in other words, please drop fd0db45546a79b6950290ced7b55314e8394ca4f). The version gets bumped when a release is made, and also, this isn't a breaking change. The new behavior is an opt-in change, so it's just a regular feature addition and gets a minor version bump as usual.

Stylistically, it would be nice if the commit messages dropped the prefixes ("refactor", "fix", "chore", etc) since it just adds extra characters and no real information that isn't already in the commit message (is this a Claude thing?).

juanpablo-santos:
Hi @nre-ableton, reverted the commit bumping version; regarding the messages, yes, it's a claude thing. I suppose that as the PRs asks about opening from topic/feature/bugfix branch it has assumed some sort of conventional commit format for the git messages

I can either rebase, rewrite and force push the new messages, or re-open a new PR with the "not so conventional commit" format, or let them be as they are right now. I was expecting the commits to be squashed so perhaps the latter is preferred, but whatever helps you most will be ok for me

nre-ableton:
> I can either rebase, rewrite and force push the new messages

Yes, that would be preferred. I'll do a more thorough code review in the meantime, but that's independent of the commit graph.

juanpablo-santos:
> > I can either rebase, rewrite and force push the new messages
> 
> Yes, that would be preferred. I'll do a more thorough code review in the meantime, but that's independent of the commit graph

Just pushed the rewritten commits, let me know if anything else is needed.

cheers,
jp

src/main/groovy/com/lesfurets/jenkins/unit/declarative/GenericPipelineDeclaration.groovy
nre-ableton: This could all be simplified to be just `componentInstance.binding = binding`, because it is no longer static.

src/main/groovy/com/lesfurets/jenkins/unit/DockerMock.groovy
```diff
@@ -3,6 +3,33 @@ package com.lesfurets.jenkins.unit
...
 class DockerMock implements Serializable {
+
+    // Helper and binding used to apply per-instance interception to the Image/Container
+    // objects this mock creates, so their calls are recorded on the owning helper's call stack.
+    // Transient: these are test-infra references and must not be carried by CPS serialization.
+    private transient PipelineTestHelper helper
+    private transient Binding binding
+
+    DockerMock() {
+    }
```
nre-ableton: This constructor should be eliminated, because then `helper` and `binding` will be `null` and probably lead to chaos in other places.

README.md
```diff
@@ -559,6 +560,36 @@ This will work fine for such a project structure:
...
+helper, and the framework no longer mutates any JVM-global state across helpers.
+
+To enable parallel execution with JUnit 5, add a `junit-platform.properties` file to your
+test resources:
+
+```properties
+junit.jupiter.execution.parallel.enabled = true
+junit.jupiter.execution.parallel.mode.default = concurrent
+junit.jupiter.execution.parallel.mode.classes.default = concurrent
+junit.jupiter.execution.parallel.config.strategy = dynamic
```
nre-ableton, line 578: But wait, aren't these already the default properties?

Should we perhaps consider setting `junit.jupiter.execution.parallel.enabled` to `false` as the default in the previous commit so that parallel testing isn't the default?

Or rather, should we keep parallel the default (which would be nice) and then change this documentation to explain how to _disable_ it if it causes problems in certain edge-cases?

nre-ableton (changes requested):
Ok, I've read through the code and have a few small suggestions, but there's nothing major that would require reworking, I don't think. 👍 

Please drop commit aaeec1386995bc0dbc52de4f3e98565010328dee, since it has nothing to do with the other changes in this PR. The `build.gradle` script needs some love to be sure, but this isn't the place to do it.

Also please drop commit dfa524a6892b9bbdd58b1d766ea10fffebe035d9 and the corresponding revert (d1e282bd07e2f68edd52993f1d790846e670efd6).

If you prefer, you can tick [the "Allow edits by maintainers" box](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) and I'll fix the git history for you.

src/main/groovy/com/lesfurets/jenkins/unit/declarative/GenericPipelineDeclaration.groovy
juanpablo-santos: had to keep the `componentInstance.hasProperty('binding')` check as not all pipelines have this property. 7 test fail if this `if` is removed

src/main/groovy/com/lesfurets/jenkins/unit/DockerMock.groovy
juanpablo-santos: done!

README.md
juanpablo-santos: unsure of what you mean here, to enable parallelism you have to provide a new properties file to the test resources folder of the project using JenkinsPipelineUnit, and provide those properties in order to enable parallel test execution.

I've reworded the section a bit, since it made reference to a 2.0.0 version, and could reword a bit more to note that the file is not provided by JenkinsPipelineUnit? Also, that doc is specific to JUnit, other test frameworks would have different configurations to enable parallel test execution.

juanpablo-santos:
oh, also, "Allow edits by mantainers" check is enabled, please feel free to do whatever you see fit on the PR :-)

README.md
nre-ableton: Well, in 8e489e83cec4242f5501789b761e0a622dba0376, you add `junit-platform.properties`, and it contains `junit.jupiter.execution.parallel.enabled = true`. So wouldn't parallel execution therefore be enabled by default unless something overrides this?

Is this properties file actually used by JUnit? Or is it just an example to be copied somewhere? I'm not terribly familiar with JUnit's behavior with respect to these files.

nre-ableton:
> oh, also, "Allow edits by mantainers" check is enabled, please feel free to do whatever you see fit on the PR :-)

Great, thanks. I just force-pushed to your branch to rewrite some history. The only thing we need to resolve now is the JUnit properties file.

README.md
juanpablo-santos: `junit-platform.properties` is added, but under `src/test/resources`, it shouldn't get included into the main jar. It just enables parallel testing on JenkinsPipelineUnit itself. 

This cuts the building time on my machine by almost 50%, which is anecdotic, as it something like 10ish seconds with parallel, ~20 without, but more importantly, should catch regressions related to parallel execution if they appear later on. 

(As a side note, I've just pushed another tiny commit, as the file added a referenced 2.0.0 release notes)

nre-ableton: Ok, great, thanks for the clarification. I will clean up the history a bit and merge this PR. Thanks for your work!

nre-ableton (approved):

nre-ableton:
I'll make a release with this patch next week.

